PR #5, opened 2023-01-31
Upgrade el8/changes

widhalmt:
Minor fixes for problems that were found during tests of the collection (most were previously undetected in the separate roles). And fixes for a stable handling of Elastic Stack version 8.

ToDos
---------

Here are some items we discovered during testing. They need to be addressed in one way or another. Either fix them here or create an issue to fix it. (Tick them off, when a fix is in this branch, in `main` or when an issue is open)

- [x] @pdolinic  / @afeefghannam89 Critical: Make sure the users /tmp directory is cleared of the old .p12 & cert file, before another rollout - Idea: Delete files in local `/tmp` after copying to target host. Will increase traffic, but make overall system more secure and stable for rebuilds. Alternatively create a unique temp directory (e.g. with `mktemp`) and use that. Partially focused in #35 
- [x] @afeefghannam89  Critical: Certificates for TLS between Beats and Logstash seem to be broken. We need a fix for that. See comments for error message. -> #40 
- [x] Investigate behavior of IP vs Hostnames in inventory. - Idea: ip address, arbitrary inventory name, hostname in inventory etc.
- [x] Optional: Kibana Web default Certificate - Idea: Use netways.ca role or have a small autogenerated certificate as default. Don't break different scenarios like company CA, local CA, letsencrypt etc. -> #39 
- [x] @pdolinic Optinal: Default Password for PleaseChangeMe - Idea: Create random passwords for everything where a default password is set and find a way so securly show it to the user. E.g. put them in a file with restricted access. -> #38
- [x] Handle hosts with more than one IP address. Currently there are hardcoded addresses chosen from Ansible facts and you can't choose which to use. This can (and did) lead to broken certificates. -> #22 
- [x] Allow for disabling TLS connection between Beats and Logstash. It regularly gave problems in the past (due to bugs in upstream code) and it would help at least with developing and testing.
- [x] Disable cluster discovery for single node clusters >= 8. There's a warning in the logs we could suppress with that -> #8 / #9
- [x] Fix system user for Kibana to access Elasticseach -> #20 
- [x] The `logstash_writer` user seems to be created but is not allowed to access Elasticsearch
- [x] Make sure to use automatic security of Elastic Stack 8 if that makes sense. We shouldn't pervert ES8 to run like ES7. ES 8 still allows to set everything manuall. Automatic security is only provided for convenience.
- [x] @widhalmt Introduce checks for version 8 (and keep them for version 7) - see https://github.com/NETWAYS/ansible-collection-elasticstack/pull/5/commits/8e447816f44590b6d09e61616c9efcffa718b38d for an example
- [x] Set default of `elastic_version` to `8`
- [x] Enable "full stack mode" by default. But allow to deactivate (e.g. for OpenSearch based installations)
- [x] Test if security is really a must with version 8 -> https://github.com/NETWAYS/ansible-collection-elasticstack/pull/5#issuecomment-1414045192 -> https://www.elastic.co/guide/en/elasticsearch/reference/current/configuring-stack-security.html#configuring-stack-security you can explicitly deactivate security even in 8. That's only interesting for testing, though.
- [x] List, which passwords can be changed by Ansible after initial setup and which can't.

Tests
--------

Every test is run with Elastic Stack version 8. `oss` variant is not supported in versions > `7` so no tests are neccessary.

- [ ] Single node with default variables (as much as possible) on Ubuntu 22.04 LTS
- [ ] Single node with default variables (as much as possible) on RedHat 8 (or similar)
- [ ] Multinode cluster with default variables (as much as possible) on Ubuntu 22.04 LTS
- [ ] Multinode cluster with default variables (as much as possible) on RedHat 8 (or similar)

### DIFF
--- a/roles/elasticsearch/templates/elasticsearch.yml.j2
+++ b/roles/elasticsearch/templates/elasticsearch.yml.j2
@@ -13,7 +13,7 @@ cluster.initial_master_nodes: [ {% for host in groups['elasticsearch']  %}
 {% if elastic_temperature is defined %}
 node.attr.temp: "{{ elastic_temperature }}"
 {% endif %}
-{% if elasticsearch_disable_systemcallfilterchecks | bool %}
+{% if elasticsearch_disable_systemcallfilterchecks | bool and elastic_release == 7 %}
 bootstrap.system_call_filter: false
 {% endif %}
 {% if elastic_variant == "elastic" %}


widhalmt:
I disabled a check for enabled security in https://github.com/NETWAYS/ansible-collection-elasticstack/pull/5/files#diff-4001bacb15c0da113993bf42c647ba9b0445d698d6a2b40733e0fbf2ed1c2397L2-R8 . Honestly, I don't know if this just works by accident, maybe only with single node clusters. It did work in my tests but we need to test more thoroughly before removing the code.

widhalmt:
I added a new milestone: https://github.com/NETWAYS/ansible-collection-elasticstack/milestone/4 . We can use it to keep track of things we stumble upon that only work with version `7`. Thanks @DanOPT for the contribution.

afeefghannam89 (approved):

PR #28, opened 2023-02-02
update kibana_system username for EL8

### LINKED ISSUE #20: "kibana" user doesn't work anymore
We fetch the password of user `kibana` for connections to Elasticsearch. The user is deprecated and in test scenarios even `kibana_system` doesn't work. So we need to find a way of getting a working user. Either figure out why `kibana_system` didn't work or create our own user via API and use that one.

pdolinic:
Change the default username from kibana to kibana_system for EL8 Upgrade

Fixes #20

### DIFF
--- a/roles/kibana/tasks/kibana-security.yml
+++ b/roles/kibana/tasks/kibana-security.yml
@@ -85,7 +85,7 @@
 
 - name: fetch Kibana password
   shell: >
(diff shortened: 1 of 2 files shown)


widhalmt (approved):
Looks good to me. Thanks for the contribution. Now merge at will.

Issue #27, opened 2023-02-02
pam_limits setting has several issues


widhalmt:
In role `elasticsearch` we set limits for Elasticsearch. This should be done by the Elasticsearch package, but didn't work in the past. I have to confess, I didn't track if works with current versions.

So this issue covers several parts:

* Changing `pam_limits` might not be desirable for every setup. We should allow for deactivating this task. The default depends on the following items
* Maybe it's not neccessary anymore. If not, we can either remove the task completely or leave it but set the default to `false`
* If it's still needed (At least in some setups) we should keep it and add a variable to deactivate it if you really don't want it
* The module `pam_limits` is part of the `community.general` collection which is not available in every kind of installation of Ansible. We need to add it to requirements. In Readme as well as in our Molecule tests.

PR #30, opened 2023-02-02
Make sure all certificates contain used hostnames

### LINKED ISSUE #14: Make sure connection string fits certificate
This can be best explained by example: In `filebeat.yml` we set the receiving hosts to a fact of all Logstash hosts (e.g. `ansible.fqdn`). We need to make sure that the same fact is part of the DN or SAN of all certificates on the receiving end.

This will make certificate creation way more robust for different setups

widhalmt:
fixes #14
needs #27 fixed to work

* Use same set of `ip`/`dns` combo for all certificates
* Add `inventory_hostname` to SAN of all certificates


widhalmt:
Needs #29 to be merged before Checks will succeed.

pdolinic (approved):

Issue #35, opened 2023-02-03
Make sure certificate handling is idempotent and reacts to changes


widhalmt:
We're creating a lot of certificates via different ways. And we're copying them via different hosts. We need to make very sure that we have a stable and reliable way to recreate certificates and even the whole CA. We need all versions: Just one certificate, all files for one host, whole CA.

Make especially sure that creating different formats like pkcs8 for Logstash is not broken because e.g. we only check for presence of the file when deciding whether we need to recreate it or not.

There are a few ideas about how to handle this:
* Deleteing temporary files right after transport
* Using unique temporary directories or filenames which won't be reused in the next run
* Checking whether the generated targets still match the source

widhalmt:
Deleting files after transport and unique temporary directories might break idempotence tests. Seems like checking is the only way to go. If you find better ways: They are very welcome.

Maybe don't check the contents of the files but use timestamps with `stat`?

afeefghannam89:
I agree with you Thomas. I have tested the collection more times, but I did not come on the same error, which you both noticed. I need really more input about the error to be able to handle this issue. As I understood, copy and fetch modules do not work as they should do. They do not update the files on target/destination when the file do not match the source.

When his right, then we should open an issue by the maintainers of these modules, because their job is to update the file on the target host.

widhalmt:
I'm sorry to be a bit misleading. In many of our tests, the collection exactly worked as expected. It's more about rethinking the whole copy and move process to make sure it really works under all conditions. Especially when you want to renew certificates.

e.g. make sure that the PKCS8 version of the Logstash key is replaced when we replace the default variant.

PR #36, opened 2023-02-03
This PR adds 2 possible Variants on Cleanup on the Controller

pdolinic:
This PR adds 2 possible Variants on Cleanup on the Controller for Stripping Secrets o
ff the /tmp directory.

1) matches .p12 & .crt in the /tmp directory without ansible_hostname
2) shell one matches the ansible_hostname."

- It is possible 1) has a very easy way to add the "ansible hostname" I haven't found.
- This PR adds into the kiba-asecurity.yml on the very end, where all the certificates have been shared, because *kibana* is expectedly on the end of the ´elasticsearch.yml´ playbook.

Ideas for Optimizations:
1) Adapt Ansible Hostname into solution Number 1
2) Create an own task file, such as `controller-cleanup.yml` in the very last role to get triggered, or allow users to switch when the cleanup will happen, themselves, via a variable.

Issues #35

### DIFF
--- a/roles/kibana/tasks/kibana-security.yml
+++ b/roles/kibana/tasks/kibana-security.yml
@@ -111,3 +111,40 @@
     Restart Kibana
   tags:
     - certificates
+
+# Version 1 with Globbing, Find and File
+
+# Cleanup ca.crt
+- name: Controller tmp cleanup of CA
+  connection: local
+  become: no
+  file:
+    path: /tmp/ca.crt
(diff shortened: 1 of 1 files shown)


widhalmt:
I'm sorry to tell you that the tasks in this change are already outdatet. There's a more sophisticated (and possibly more accurate and faster) version in `main` now. So I'm sorry I have to tell you that I'm closing this PR now.

PR #44, opened 2023-02-06
Use verify from cluster 8

### LINKED ISSUE #37: Add "verify.yml" to cluster scenarios
We want at least to know if the hosts are connected to each other. Remember that `elasticsearch_no-security` is also clustered in some implementations.

widhalmt:
fixes #37


afeefghannam89 (changes requested):
Do you want really to not run commands while running Ansible in check mode by using **ignore_errors: "{{ ansible_check_mode }}"**?
Maybe I am wrong, as I see you use the registered output of this tasks in the next tasks, maybe **check_mode: no** will be better? With this parameter the task will be run in both check and normal mode and register his output to be used later from other tasks.

widhalmt:
> Do you want really to suppress commands errors while running Ansible in check mode by using **ignore_errors: "{{ ansible_check_mode }}"**? Maybe I am wrong, as I see you use the registered output of this tasks in the next tasks, maybe **check_mode: no** will be better? With this parameter the task will be run in both check and normal mode and register his output to be used later from other tasks.

You mean in https://github.com/NETWAYS/ansible-collection-elasticstack/pull/44/files#file-roles-elasticsearch-tasks-elasticsearch-security-yml-L116 ? That's in an unchanged file. I just moved `verify.yml` to another molecule scenario in this PR.

The code you're referring to is from an external contribution. I'll have a second look into it and if we need to change it, I'll open another issue. Thanks for noticing!

afeefghannam89 (approved):
Ok check, I will approve then this change while you will check the sense of this parameter later. Thanks Thomas :)

PR #46, opened 2023-02-07
Fix #40

### LINKED ISSUE #40: Fix TLS certificates for Beats connection
In some setups, Logstash can't read it's own PKCS8 formatted certificates for `beats` input. It seems, like some have problems with encrypted certificates or password protected certificates. The weird part is that some setups work completely and others break.

We need a reliable method to create certificates as secure as possible without breaking.

afeefghannam89:
This fix will let Logstash be able to read the certificate key file.

Fix #40


martialblog (changes requested):
As discussed, I would change the Commit message to be more descriptive

widhalmt:
We tested an abundance of combinations of Logstash, Beats-Plugin and JREs and all we could verify is that "it depends". Some work, others don't but we couldn't find a pattern that will reliably work. So guess, the next steps are to deactivate encryption of certificates and add an issue to retest in a few verions.

widhalmt (approved):
It looks like the failed test only happened due to a timeout. Hereby I approve.

PR #55, opened 2023-02-10
Add fileset for system module

### LINKED ISSUE #54: Enabling modules in version 8 needs filesets
See https://github.com/NETWAYS/ansible-collection-elasticstack/actions/runs/4143459524/jobs/7165345423#step:6:614 all we have so far. Please dig into the topic and provide a fix.

widhalmt:
fixes #54

Beginning with release version 8 you need to give a fileset for "system" module of filebeat. Version 7 just used defaults.

Since we are thinking about replacing modules by elastic agents and filebeat only without modules in special cases, I add a quick fix and flag it as experimental. Let's see how it works.

### DIFF
--- a/roles/beats/tasks/filebeat.yml
+++ b/roles/beats/tasks/filebeat.yml
@@ -37,28 +37,40 @@
     - filebeat_configuration
     - beats_configuration
 
-- name: Enable modules
-  command: "filebeat modules enable {{ item }}"
-  args:
-    creates: "/etc/filebeat/modules.d/{{ item }}.yml"
-  with_items: "{{ filebeat_modules }}"
+- name: Configure modules
   when: filebeat_modules is defined
+  with_items: "{{ filebeat_modules }}"
   tags:
     - configuration
     - filebeat_configuration
     - beats_configuration
+  block:
 
-- name: Enable Ingest Pipelines
-  command: >
-    /usr/bin/filebeat setup --pipelines --modules {{ item }} &&
-    /usr/bin/filebeat version > /etc/filebeat/{{ item }}_pipeline_created
-  args:
-    creates: "/etc/filebeat/{{ item }}_pipeline_created"
-  notify:
-    - Restart Filebeat
-  changed_when: false
-  with_items: "{{ filebeat_modules }}"
-  when: filebeat_modules is defined
+    - name: Enable modules
+      command: "filebeat modules enable {{ item }}"
+      args:
+        creates: "/etc/filebeat/modules.d/{{ item }}.yml"
+
+    - name: Enable Ingest Pipelines
+      command: >
+        /usr/bin/filebeat setup --pipelines --modules {{ item }} &&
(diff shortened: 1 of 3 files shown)


widhalmt:
My initial diagnosis was wrong: https://discuss.elastic.co/t/filebeat-setup-exiting-module-system-is-configured-but-has-no-enabled-filesets/311584 . We just need to change the way we setup modules. Please note that #77 could make using modules in filebeat completely obsolete. Therefore I didn't put much effort in clean code for enabling modules. If we want to continue to support, we should find a better way than just running the same code over and over again.

widhalmt:
Default configuration with added modules will connect Filebeat to Logstash and then run `filebeat setup`. For all that I know, this won't work. (Even when it did in a local test). So either don't allow that combination or skip `setup` when talking to Logstash.

Or, the most sophisticated solution: Make sure, Filebeat talks to Elasticsearch when doing `setup` and then connect to Logstash. This could help with working with differently configured beats. But it would need more thinking than what I have time for now.

pdolinic (dismissed):

roles/beats/tasks/filebeat.yml
```diff
@@ -37,28 +37,40 @@
...
+    - name: Enable Ingest Pipelines
+      command: >
+        /usr/bin/filebeat setup --pipelines &&
+        /usr/bin/filebeat version > /etc/filebeat/{{ item }}_pipeline_created
+      args:
+        creates: "/etc/filebeat/{{ item }}_pipeline_created"
+      with_items: "{{ filebeat_modules }}"
+      notify:
+        - Restart Filebeat
+      changed_when: false
```
afeefghannam89, line 73: I am annoying, I know :-) why should Ansible reporting no changes?

afeefghannam89 (dismissed):

roles/beats/tasks/filebeat.yml
widhalmt: Because it's a shell command that's executed with every run. It's quite hard to determine if this very command is still neccessary or not. But the command itself is idempotent and you can re-run it every time. You're right, it would be way more elegant to check if we need to run it or not, but I couldn't find a reliable way to check that's easy enough to implement.

afeefghannam89: I would say you could merge this, and we open an issue to evaluate the idempotent principle for the collection later, because it is not easy and it would take time what do you think?

afeefghannam89 (approved):

roles/beats/tasks/filebeat.yml
afeefghannam89: You can close this conversation after you see this suggestion :)

widhalmt: Yes, thanks. I guess, this is an even bigger item when we take #77 into account. Maybe we can remove the whole modules part.

DanOPT (approved):
Looks good to me 👍

Issue #57, opened 2023-02-12
Kibana does not start after reboot


afeefghannam89:
Kibana service should start after reboot

widhalmt:
That was already reported in the old role: https://github.com/NETWAYS/ansible-collection-elasticstack/issues/69 Sorry for not copying over the issue sooner.

Issue #56, opened 2023-02-12
kibana uses the same certificate from elasticsearch


afeefghannam89:
kibana uses the same key of elasticsearch. When we want to change the key of one of them, then cluster will not work.

PR #58, opened 2023-02-13
Enable ca/cert management

### LINKED ISSUE #35: Make sure certificate handling is idempotent and reacts to changes

afeefghannam89:
* Enable renew CA
* Enable renew Elasticsearch cert
* Enable renew Logstash cert
* Enable renew Kibana cert
* Enable renew beats cert
* Check CA and Certs validity date and renew them 
* Let Logstash write data on disk
* Prevent Logstash to stick by Elasticsearch dead connection
* Make sure certificate handling is idempotent and reacts to changes #35
* Make Kibana use its certificate and not share the on from elasticsearch #56
* Let Kibana start after reboot #57 #69
* Allow to use persisted queues per pipeline #60
* Restart the available beat service when create or renew certificates #83


roles/elasticsearch/tasks/elasticsearch-security.yml
```diff
@@ -244,8 +387,11 @@
     mode: 0640
   notify:
     - Restart Elasticsearch
+    - Restart Kibana
```
widhalmt, line 390: This handler is not available in this role, yet. We only have it in the `kibana` role. Please keep in mind, that most setups have an arbitrary number of Kibana hosts running on different hosts.

docs/role-logstash.md
```diff
@@ -62,6 +62,9 @@ Aside from `logstash.yml` we can manage Logstashs pipelines.
 * *logstash_password*: Password of Elasticsearch user (Default: `password`)
 * *logstash_user_indices*: Indices the user has access to (Default: `'"ecs-logstash*", "logstash*", "logs*"'`)
 * *logstash_reset_writer_role*: Reset user and role with every run: (Default: `true`)
+* *logstash_validate_after_inactivity*: How long should logstash wait, before starting a new connection and leave the old one with elasticsearch, when the connection with elasticsearch get lost: (Default: `0`).
+* *logstash_queue_type*: Where should Logstash right the data: (Default: `persisted`)
```
widhalmt, line 66: How about making the other setting (`memory`?) the default? Persisted queues can have a huge impact on Logstash, creating lots of I/O and slowing down the whole process. At least pipelines that receive messages from senders which can't deal with backpressure should be able to disable persistance.

roles/beats/tasks/main.yml
```diff
@@ -2,37 +2,38 @@
...
+      - beats_ca_dir is undefined
 
-  - name: Set common password for common certificates
-    set_fact:
-      beats_tls_key_passphrase: "{{ elastic_cert_pass }}"
-    when: elastic_cert_pass is defined
+    - name: Set common password for common certificates
+      set_fact:
+        beats_tls_key_passphrase: "{{ elastic_cert_pass }}"
+      when: elastic_cert_pass is defined
```
widhalmt, line 32:
```suggestion
      when: 
        - elastic_cert_pass is defined
        - beats_tls_key_passphrase is undefined
```
I guess syncing the two is only really neccessary when you use one single certificate for both services. As I understand, you cleaned that up. So I assume, it would work to have different passwords or even unique passwords per host.

roles/elasticsearch/tasks/elasticsearch-security.yml
```diff
@@ -257,8 +403,11 @@
...
```
widhalmt, line 406: Same as above. That handler doesn't exist in this role.

roles/elasticsearch/tasks/main.yml
```diff
@@ -12,7 +12,11 @@
   set_fact:
     elasticsearch_tls_key_passphrase: "{{ elastic_cert_pass }}"
   when:
-    - elastic_cert_pass is defined
+  - elastic_cert_pass is defined
```
widhalmt, line 15:
```suggestion
    - elastic_cert_pass is defined
```
I guess 2 spaces more of indentation are needed here.

roles/logstash/defaults/main.yml
```diff
@@ -34,6 +34,8 @@ logstash_pipelines:
 #     name: shipper
 #     source: https://github.com/widhalmt/shipper-logstash-pipeline.git
 logstash_elasticsearch_output: true
+logstash_validate_after_inactivity: '0'
```
widhalmt, line 37: Does `0` work? I know, we discussed that but isn't that a bit aggressive for production use? I'm not sure myself I just want to be cautious.

roles/logstash/tasks/logstash-security.yml
```diff
@@ -135,13 +286,22 @@
     -nocrypt
   args:
     creates: "{{ logstash_certs_dir }}/{{ inventory_hostname }}-pkcs8.key"
+  when: not logstash_beats_tls_encryptkey | bool
```
widhalmt, line 289: Didn't you remove the other PKCS8 transformation? Do I read it right that, when `logstash_beats_tls_encryptkey` is set that there's no PKCS8 key generated at all?

roles/logstash/tasks/logstash-security.yml
```diff
@@ -124,6 +273,8 @@
     remote_src: true
   tags:
     - certificates
+    - renew_ca
+    - renew_logstash_cert
```
widhalmt, line 278: I guess, there's a way of moving the old PKCS8 key missing. Maybe I just missed it. But when the key is renewed completely, the PKCS8 variant won't be renewed, because the tasks only checks for the `creates`, right?

roles/logstash/templates/pipelines.yml.j2
```diff
@@ -30,6 +30,8 @@
 
 - pipeline.id: ansible-forwarder
   path.config: "/etc/logstash/conf.d/ansible-forwarder/*.conf"
+  queue.type: {{ logstash_queue_type }}
```
widhalmt, line 33: I'd love to have that configureable per pipeline and not globally. Falling back to a global default would be great, but we need a bit more flexibility than global settings. For both `queue.type` and `queue.max.bytes`.

widhalmt (changes requested):
I love how you work with tags everywhere and how you use tags for one-shot runs like when renewing certificates.

I'm not 100% sure if renewing client certificates wouldn't make sense as a variable. This way you could just enable it and have them replaced as soon as they close in on the end of their validity.

Most comments are things I'm wondering and I'm not really sure if there needs to be a change. Please don't read them as a request for an actual change. It's more of a "did you have time to test"?

The only real changes I see are that handlers are missing (we talked about that offline) and the per queue setting (with global default) of queue type and size.

roles/elasticsearch/tasks/elasticsearch-security.yml
afeefghannam89: Thanks for this good found :), I did not see it. I will make the suitable changes.

docs/role-logstash.md
afeefghannam89: What do you mean please with **other** other pipelines?

roles/logstash/tasks/logstash-security.yml
afeefghannam89: You are right, I found this after pushing and it is already in. the next push with I did not make yet. It did not catch it during master rebaise

afeefghannam89:
> I love how you work with tags everywhere and how you use tags for one-shot runs like when renewing certificates.
> 
> I'm not 100% sure if renewing client certificates wouldn't make sense as a variable. This way you could just enable it and have them replaced as soon as they close in on the end of their validity.
> 
> Most comments are things I'm wondering and I'm not really sure if there needs to be a change. Please don't read them as a request for an actual change. It's more of a "did you have time to test"?
> 
> The only real changes I see are that handlers are missing (we talked about that offline) and the per queue setting (with global default) of queue type and size.

I like tags more, because they are more flexible by management of changes, the user should not change any variable value when he would like to run a snippet of Ansible code. The variable will come later when we check the validity of certificate. This variable will be set to the conditions of certificate renew tasks. So the user does not have to manipulate any variable value, this will enforce automatism Principe

docs/role-logstash.md
widhalmt: Ususally we have more than one pipeline in a Logstash configuration and in `pipelines.yml`. I'd prefer to set `logstash_queue_type` and `logstash_queue_memory_max_bytes` different per pipeline. Not one global value for all of them.

See #60 . It's an old issue from the deprecated role but we already thought about the same thing.

roles/logstash/templates/pipelines.yml.j2
afeefghannam89: Thanks for the advice :) done

roles/logstash/defaults/main.yml
afeefghannam89: I raised it to 300 = 5 minutes

roles/beats/tasks/main.yml
afeefghannam89: Ahhh good notice, I updated it

docs/role-logstash.md
afeefghannam89: I hope it is now more flexible see please last pushes :)

roles/logstash/tasks/logstash-security.yml
afeefghannam89: The complete Logstash certificate directory will be backed up and removed, so no need to move files individually ;)

widhalmt (changes requested):
Wow, that was a big one to review. :-)

The only real change I can request is the missing "newline" character. The rest is more "I'm not sure, can you give me information about it". So I guess, it's very close to being finished.

The "boolean" for the queue type was just an idea. I guess, it's not worth editing everything again just for that.

widhalmt:
Please merge `main` into this branch. It should help with fixing some problems.

roles/logstash/templates/pipelines.yml.j2
widhalmt: Thanks. :-)

widhalmt:
I fixed some lint so checks can go through. I'm right now testing this on a local system.

widhalmt:
Changing the Logstash certificate works for me. But only for the `elasticsearch` output, not for `monitoring` Since `xpack.monitoring` was deprecated with Elastic Stack 7 anyway, I'll add a condition so that it will be removed from 8 setups.

roles/kibana/tasks/kibana-security.yml
```diff
@@ -108,6 +215,9 @@
     group: kibana
     mode: 0640
   notify:
-    Restart Kibana
+   - Restart Kibana
+   - Restart Elasticsearch for Kibana certificates
```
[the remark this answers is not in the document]
afeefghannam89: Thanks you are right I have remove it.

roles/beats/handlers/main.yml
```diff
@@ -5,16 +5,16 @@
   service:
     name: filebeat
     state: restarted
-  when: filebeat_enable | bool
+  when: beats_filebeat | bool
```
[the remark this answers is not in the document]
afeefghannam89: When the certificate will be changed, the beat services have to be restarted, right?
So when we stay with for example  *auditbeat_enable | bool* as condition and auditbeat is not installed, but it is enabled as default, then Ansible will throw an error saying that auditbeat service is not available.

docs/role-logstash.md
```diff
@@ -47,21 +47,25 @@ Aside from `logstash.yml` we can manage Logstashs pipelines.
...
+* *logstash_tls_key_passphrase*: Passphrase for Logstash certificates (default: `LogstashChangeMe`)
+* *elastic_ca_pass*: Password for Elasticsearch CA (default: `PleaseChangeMe`)
 * *logstash_elasticsearch*: Address of Elasticsearch instance for default output (default: list of Elasticsearch nodes from `elasticsearch` role or `localhost` when used standalone)
 * *logstash_security*: Enable X-Security (No default set, but will be activated when in full stack mode)
 * *logstash_user*: Name of the user to connect to Elasticsearch (Default: `logstash_writer`)
 * *logstash_password*: Password of Elasticsearch user (Default: `password`)
 * *logstash_user_indices*: Indices the user has access to (Default: `'"ecs-logstash*", "logstash*", "logs*"'`)
 * *logstash_reset_writer_role*: Reset user and role with every run: (Default: `true`)
+* *logstash_validate_after_inactivity*: How long should logstash wait, before starting a new connection and leave the old one with elasticsearch, when the connection with elasticsearch get lost: (Default: `0`).
+* *logstash_queue_type*: Where should Logstash right the data: (Default: `persisted`)
```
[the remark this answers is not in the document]
afeefghannam89: This version is not bad also :)

afeefghannam89:
@widhalmt I am fixing Lint now

widhalmt:
> @widhalmt I am fixing Lint now

We have some other Pull requests which might fix this already. Have a look, maybe it only needs a Review to be merged and then you can safe that pesky work.

roles/beats/handlers/main.yml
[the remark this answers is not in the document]
widhalmt: Right! Good catch. So we need both conditions, I guess.

docs/role-beats.md
```diff
@@ -102,7 +102,9 @@ If you want to use this role with your own TLS certificates, use these variables
 * *beats_ca_dir*: Path to custom CA certificates and keys (default: none - if not set will be filled with different values depending on which Stack variant is used)
 * *beats_tls_key*: Path to the keyfiles (default: `{{ beats_ca_dir }}/{{ ansible_hostname }}.key`)
 * *beats_tls_cert*: Path to the certificate (default: `{{ beats_ca_dir }}/{{ ansible_hostname }}.crt`)
-* *beats_tls_key_passphrase*: Passphrase of the keyfile (default: `ChangeMe`)
+* *beats_tls_key_passphrase*: Passphrase of the keyfile (default: `BeatsChangeMe`)
+* *beats_cert_expiration_buffer*: Ansible will renew the beats certificate if its validity is shorter than this value, which should be number of days. (default: 30)
+* *beats_cert_will_expire_soon*: Set it to true to renew beats certificate (default: `fasle`), Or run the playbook with `--tags renew_beats_cert` to do that.
```
widhalmt, line 107:
```suggestion
* *beats_cert_will_expire_soon*: Set it to true to renew beats certificate (default: `false`), Or run the playbook with `--tags renew_beats_cert` to do that.
```

roles/elasticsearch/defaults/main.yml
```diff
@@ -25,7 +25,11 @@ elasticsearch_jna_workaround: false
 elastic_ca_dir: /opt/es-ca
 elastic_initial_passwords: /usr/share/elasticsearch/initial_passwords
 elastic_ca_pass: PleaseChangeMe
-elastic_cert_pass: PleaseChangeMeIndividually
+elasticsearch_tls_key_passphrase: PleaseChangeMeIndividually
+ca_expiration_buffer: 30
```
widhalmt, line 29:
```suggestion
elastic_ca_expiration_buffer: 30
```

Please keep the prefix. Otherwise you might have collisions with our `ca` role

roles/elasticsearch/defaults/main.yml
```diff
@@ -25,7 +25,11 @@ elasticsearch_jna_workaround: false
...
+elasticsearch_cert_expiration_buffer: 30
+ca_will_expire_soon: false
```
widhalmt, line 31:
```suggestion
elastic_ca_will_expire_soon: false
```

Please keep the prefix. Otherwise you might have collisions with our `ca` role

widhalmt (changes requested):
I just found one typo and 2 missing prefixes for variable names. Everything else looks not just good but excellent to me!

roles/elasticsearch/defaults/main.yml
afeefghannam89: Done.

roles/elasticsearch/defaults/main.yml
afeefghannam89: Done.

docs/role-beats.md
afeefghannam89: Done

roles/beats/handlers/main.yml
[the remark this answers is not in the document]
afeefghannam89: Done.

PR #88, opened 2023-02-15
Lint full codebase before merging

widhalmt:
I'm using this mostly to test mandatory status checks


widhalmt:
The idea of this PR is:

* Have a linter run over the whole collection for every PR against `main`
* Have a molecule test with `rockylinux8` and `ubuntu2204` for the whole stack for every PR against `main`
* Have more detailed molecule checks per role only when the roles code is changed in this PR (Thanks @mkayontour for that code!)
* Have a full cluster check *after* every merge to `main` for every supported OS

I'm still struggling a bit with GitHub's Action syntax.

widhalmt:
It looks like the tests for changes in roles only run once when a push to their code is made. They don't run, when the PR will change them. Seems like I just learned the difference between the `push` and `pull_request` triggers. 🤣  🤦🏽 

I'm not sure which trigger I like better. Any opinions?

mkayontour (dismissed):

widhalmt:
Since some checks will only run when adding the PR to the queue, we need a quick approve, even when the approver didn't have time to dig into the code as a whole. This is an exception, hopefully and applies to all approves after this comment.

mocdaniel (approved):

PR #101, opened 2023-02-17
Install a specific version of all tools

### LINKED ISSUE #76: Optionally install specific version of Beats
A possible solution could be similar to what we did to Logstash in https://github.com/NETWAYS/ansible-role-logstash/pull/75

Two notes, though:
1. If you find a better way: You're very welcome. The ugly string you have to write to use a specific version was chosen to have a solution that will work with `rpm` and `deb` versions. Maybe there's a better way.
2. You should add installation of a specific version to Molecule. If you do so, add a test to `verify.yml` to make sure the right version is actually installed.

### LINKED ISSUE #98: Allow for installation of a specific version of tools
Allow for installing a specific version of all tools.

Two things to note:
* Allowing for different versions within a stack makes no sense at all. So we need a common variable to tell Ansible which version to install
* Upgrading existing systems is way beyond the scope of this issue! There are plans for automated updates in the future. Right now you're chaning the version information at your own risk! We need to make this absolutely clear in the Readme

widhalmt:
We now have a global variable that allows installatin of a specific version of all tools. Since it doesn't make sense to have different versions in a single stack, we don't allow different settings for it.

Updating is not part of this pull request.

fixes #76 
fixes #98


widhalmt:
Linter findings are addressed in #58

widhalmt:
We should make selecting a version easier. Like build the complete package name with an OS dependent divider we set automatically. This would allow to just add the version numer as a variable without `-` or `=`.

And we should reactivate `verify` based tests. Only activate them when the matching release is set and use some that are new enough to be in the repositories.

widhalmt:
This branch is a dead end. It's code is superseded by that in #108 . We need a totally different approach for installing different versions, releases etc. when allowing to disable repos. So the code in here is obsolete.

PR #108, opened 2023-02-20
Allow disabling repos per default

### LINKED ISSUE #76: Optionally install specific version of Beats

### LINKED ISSUE #98: Allow for installation of a specific version of tools

### LINKED ISSUE #106: Elastic Repos
Hi,

Elastic Repos should be disabled by default (enabled=0), so that it is not updated with os updates.

In each roles at point installation they should be enabled explicit with enablerepo parameter.

widhalmt:
This introduces a new variable that will disable yum repositories per default and re-enables them temporarily when modifying packages. It's not possible to disable apt repositories temporarily so we need to have two different approaches.

fixes #106
fixes #76
fixes #98
supersedes #101


widhalmt:
Looks like it's like I was fearing when I started this PR. We'll need different tasks for rpm and deb based installations. This lets the counts of installation tasks explode because we already have to take care of several special situations. Maybe we can come up with a simpler solution.

afeefghannam89 (approved):
Whoop, this will make our life easier thanks

PR #113, opened 2023-02-21
Failsafe for standalone usage

### LINKED ISSUE #112: Roles in standalone mode rely on other roles for certificate management
When you run roles on their own ( `elastic_stack_full_stack: false` ) the newly introduced certificate handling tasks and handling still try to access hosts handled by other roles. We need `when` clauses to prohibit that.

widhalmt:
fixes #112

The roles of this community can either be used together relying on tasks and facts of each other or standalone. We need to make sure that no task that runs in standalone mode tries to access information from other roles.


widhalmt:
Needs #108 for some checks to run through. Required checks should work, though.

PR #115, opened 2023-02-24
Replace hardcoded 9200 with variable

### LINKED ISSUE #70: Set Elasticsearch port via variable
We now use hardcoded port `9200` for Elasticsearch connections. We should allow for setting it via variable

widhalmt:
fixes #70

### DIFF
--- a/molecule/kibana_full_stack/verify.yml
+++ b/molecule/kibana_full_stack/verify.yml
@@ -7,8 +7,6 @@
     elastic_elasticsearch_http_port: 9200
 
   tasks:
-  vars:
-    elastic_elasticsearch_http_port: 9200
 
   - name: Give some time for all services to connect
     wait_for:
--- a/molecule/kibana_full_stack/verify.yml
+++ b/molecule/kibana_full_stack/verify.yml
@@ -3,51 +3,53 @@
 
 - name: Verify
   hosts: all
+  vars:
+    elastic_elasticsearch_http_port: 9200
+
   tasks:
-
-  - name: Give some time for all services to connect
-    wait_for:
-      timeout: 120
-
-  - name: Kibana checks
-    block:
-
-    - name: fetch kibana.yml
-      command: cat /etc/kibana/kibana.yml
-      register: kibanayml
-
-    - name: Show kibana.yml
-      debug:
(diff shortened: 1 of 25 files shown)


molecule/kibana_full_stack/verify.yml
```diff
@@ -3,51 +3,55 @@
...
-  #  register: kibana_status
-  #  #failed_when: "'"title": "Green"' not in kibana_status.content"
-  #  failed_when: "'Green' not in kibana_status.content"
+      - name: fetch kibana.yml
+        command: cat /etc/kibana/kibana.yml
+        register: kibanayml
+
+      - name: Show kibana.yml
+        debug:
+          var: kibanayml.stdout_lines
```
mocdaniel, line 25: Is this debug blog meant to end up in the PR?

molecule/kibana_full_stack/verify.yml
```diff
@@ -3,51 +3,55 @@
...
+    # The following might be nicer but doesn't work
+    #- name: Connect to Kibana
+    #  uri:
+    #    url: http://ansible-role-kibana_full_stack:5601/api/status
+    #    user: elastic
+    #    password: "{{ elastic_password.stdout }}"
+    #    return_content: yes
+    #  register: kibana_status
+    #  #failed_when: "'"title": "Green"' not in kibana_status.content"
+    #  failed_when: "'Green' not in kibana_status.content"
```
mocdaniel, line 57: If left in here as comment, maybe add *why* it's not working, as this might change in the future.

mocdaniel (approved):
Two minor questions, but looks good overall.

molecule/kibana_full_stack/verify.yml
widhalmt: The task just failed. :-D I didn't more of investigating. Linters kept telling me that this would be better and I believe them, but I couldn't get it to work in time.

molecule/kibana_full_stack/verify.yml
widhalmt: Yes. It's only in the Molecule `verify.yml` and it helped a lot with checking if templating went right. Good catch, though.

Issue #84, opened 2023-02-14
Change hard coded values to variables


afeefghannam89:
We should change the hard coded lines/values to variable in the roles, mostly in beats.

PR #116, opened 2023-02-24
Set beats port via global variable

### LINKED ISSUE #78: Rename "ports" variables to work globally with all our roles
We lately introduced variables to set ports of receiving Elasticsearch and Logstash hosts. We need to rename them so all roles can work together to set the ports accordingly.

widhalmt:
fixes #78

We had different approaches to set the port the beats protocol uses throughout our roles. Now we have one single variable.


widhalmt:
Need https://github.com/NETWAYS/ansible-collection-elasticstack/pull/113 to be merged before checks can run through

widhalmt:
helps with #84

DanOPT (approved):
Looks good!

PR #122, opened 2023-03-14
Rework automated tests

### LINKED ISSUE #120: Rework testing scenarios
When moving from separate roles to a collection we tried to integrate existing tests as much as we could in the new collection.

Telling from some experience with the new collection, we might have a few holes but more important, we have redundant and very thorough tests. Long running (and failing due to timeout) tests make development a lot more cumbersome than it has to be.

So we should definitelty rework how, when and what we test.

widhalmt:
We have lots and lots of tests that stem from the elastic stack roles we had before this collection. While developing the collection further we found that some tests are now redundant, others are useless now and even more others are missing.

So we take the chance to rework our automated checks as much as neccessary to get consistent checks that won't take forever to run for a simple small change.

fixes #120
fixes #122 through the remove of molecule full stack scenarios in beats, logstash and kibana related GitHub workflows.


afeefghannam89 (approved):
I like the verify tasks.

PR #125, opened 2023-03-20
Make cert expiration check independent of locale

### LINKED ISSUE #124: Calculation of certificate expiration date won't work with some locales
When running the collection a second time over a test system I get an error when it tries to calculate the expiration date of certificates.

```
TASK [netways.elasticstack.elasticsearch : Set the ca expiration date in days] **************************************************************************************************************************************************************
fatal: [elastic801]: FAILED! => {"msg": "the field 'args' has an invalid value ({'elastic_ca_expiration_days': \"{{ ((ca_expiration_date.stdout | to_datetime('%b %d %H:%M:%S %Y %Z')) - (ansible_date_time.date | to_datetime('%Y-%m-%d'))).days }}\"}), and could not be converted to an dict.The error was: time data 'Mar 16 15:55:01 2026 GMT' does not match format '%b %d %H:%M:%S %Y %Z'\n\nThe error appears to be in '/home/widhalmt/.ansible/collections/ansible_collections/netways/elasticstack/roles/elasticsearch/tasks/elasticsearch-security.yml': line 34, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: Set the ca expiration date in days\n  ^ here\n"}
```

The `to_datetime` calculation is correct and the string *does* match the pattern. Only problem is that `%b` refers to "Month as locale’s abbreviated name." . So this can or can not work depending on locales on all involved machines.

widhalmt:
This will use another module for checking expiration dates of certificates

fixes #124

### DIFF
--- a/roles/elasticsearch/tasks/elasticsearch-security.yml
+++ b/roles/elasticsearch/tasks/elasticsearch-security.yml
@@ -15,31 +15,37 @@
   register: ca_exists
   when: inventory_hostname == elasticsearch_ca
 
-- name: Get the ca expiration date # noqa: risky-shell-pipe
-  shell: >-
-    if test -v BASH; then set -o pipefail; fi;
+# This pkcs12 store contains the CA
+# Therefore we need to extract the certificate with clcerts
+# not the CA cert with cacerts
+
+- name: Get CA certificate in PEM format
+  command: >-
     openssl pkcs12
     -in '{{ elastic_ca_dir }}/elastic-stack-ca.p12'
-    -nodes
-    -passin pass:'{{ elastic_ca_pass }}' |
-    openssl x509 -noout -enddate |
-    awk -F'=' '{print $2}'
-  register: ca_expiration_date
-  args:
-    executable: /bin/bash
+    -passin pass:'{{ elastic_ca_pass }}'
+    -clcerts
+    -nokeys
+  register: ca_pem
   changed_when: false
-  no_log: true
   when: inventory_hostname == elasticsearch_ca and ca_exists.stat.exists | bool
 
-- name: Set the ca expiration date in days
-  set_fact:
-    elastic_ca_expiration_days: "{{ ((ca_expiration_date.stdout | to_datetime('%b %d %H:%M:%S %Y %Z')) - (ansible_date_time.date | to_datetime('%Y-%m-%d'))).days }}"
-  when: inventory_hostname == elasticsearch_ca and ca_expiration_date.skipped is not defined
+- name: Check CA expiration date
+  community.crypto.x509_certificate_info:
+    content: "{{ ca_pem.stdout }}"
+    valid_at:
+      expiration: "+{{ elastic_ca_expiration_buffer | int }}d"
+  register: ca_info
+  when:
+    - inventory_hostname == elasticsearch_ca
+    - ca_exists.stat.exists | bool
 
 - name: Set ca will expire soon to true
   set_fact:
     elastic_ca_will_expire_soon: true
-  when: inventory_hostname == elasticsearch_ca and elastic_ca_expiration_days is defined and elastic_ca_expiration_days | int <= elastic_ca_expiration_buffer | int
+  when:
+    - inventory_hostname == elasticsearch_ca
+    - not ca_info.valid_at.expiration | bool
 
 - name: Print the ca renew message
   debug:


widhalmt:
Please wait with reviewing since I could only change the checks for the CA cert for now.

widhalmt:
@DanOPT you found an alternative way to deal with the locale in the function. Let's try that because it needs less changes than my approach.

DanOPT:
Uploaded the filter plugin.

Tested with:
  - Python 2.7
  - Python 3.10
  - pep8 (Sanity test)

We should implement a new workflow in Actions for the modules. More than one workflow is probably not required.

See more here for requirements of the workflow (like pep8 and for filter plugins Python 2.6 and Python 3.5:
- https://docs.ansible.com/ansible/latest/dev_guide/developing_modules_checklist.html

DanOPT:
**Changes**
- No changes in original tasks, just added a fallback
- Introduced filter `netways.elasticstack.es_to_date()`
- Introduced new pipeline in Actions to test the new module (will only be triggered in fullstack workflow and if the code of plugins/** and other relevant paths has been changed)

**netways.elasticstack.es_to_date filter**
This is the explanation for the naming of all variables that are possible parameters in the filter:

`string`:
This is the original name of the variable in the core. **Default**: undefined
Link: https://github.com/ansible/ansible/blob/devel/lib/ansible/plugins/filter/core.py#L94)

`format`:
This is the original name of the variable in the core **Default:** "%Y-%m-%d %H:%M:%S"
Link: https://github.com/ansible/ansible/blob/devel/lib/ansible/plugins/filter/core.py#L94)

`language_code`:
The language code is part of the _sequence_ that locale.getlocale() returns. 
Further reading in the documentation of the function: https://docs.python.org/3/library/locale.html#locale.getlocale

`encoding`:
The language code is part of the _sequence_ that locale.getlocale() returns. 
Further reading in the documentation of the function: https://docs.python.org/3/library/locale.html#locale.getlocale

`sequence`:
The `sequence` variable in the filter saves the return of locale.getlocale() which is a tuple e.g.: ( 'en_US', 'UTF-8' )

**Workflow test_plugins.yml**
- Will only be triggered if a file of relevant paths has been changed and in the test_full_stack.yml

**Thoughts**
- We can now start reducing line-length. Unfortunately I have not enough time right now, but I will tackle it.
- It would be a better solution to write a module and execute the function directly on the remote node. That will massively reduce tasks. I'm sure about that because I have a little bit experience with standard modules.
- I will need to implement pep8 and Python 2.7 and 3.7 tests. I will tackle that tomorrow. As soon as that is implemented we can test all modules we will create and have a workflow for future projects.
- I included doc_fragments in the Plugin in case we get the documentation automated e.g. with antsibull
- Now I'm able to implement modules much faster, because I already made my first steps now.

DanOPT:
This is a better solution. The module is called ca_info. If the naming of something is not appropriate, feel free to change it.

I tested the test_plugins.yml workflow, unfortunately, I didn't have the time to test the Role to see if everything is working. I also didn't have the time to document it, but I will do it as soon as possible. If you want to merge we can also do the documentation in a separate PR.

That's the premium solution. It reduced line-length, tasks, and readability. Hope you like it 😃

DanOPT:
**Parameters**
`ca_dir`: The directory of the CA certificate. **Required**
`password`: The password of the pkcs12 file. **Default:** `None`

**Example**
```
- name: Test
  netways.elasticstack.ca_info:
    ca_dir: /home/daniel/es-ca/
    password: PleaseChangeMe
  register: ca_infos
- name: Debug
  debug:
    msg: "{{ ca_infos }}"
```

**Returns**
```
{
"changed": false,
"extensions": {
    "authorityKeyIdentifier": {
        "critical": "False",
        "oid": "2.5.29.35,",
        "values": {
            "authority_cert_issuer": "None",
            "authority_cert_serial_number": "None",
            "key_identifier": "b'\\x82S \\x11\\xc7s\\xa7^*w\\xc1\\xdf\"\\xe4#\\xb4\\xc4P\\xba\\xcf'"
        }
    },
    "basicConstraints": {
        "critical": "True",
        "oid": "2.5.29.19,",
        "values": {
            "ca": "True",
            "path_length": "None"
        }
    },
    "subjectKeyIdentifier": {
        "critical": "False",
        "oid": "2.5.29.14,",
        "values": {
            "digest": "b'\\x82S \\x11\\xc7s\\xa7^*w\\xc1\\xdf\"\\xe4#\\xb4\\xc4P\\xba\\xcf'"
        }
    }
},
"failed": false,
"issuer": "Elastic Certificate Tool Autogenerated CA",
"not_valid_after": "2026-03-28 01:58:02",
"not_valid_before": "2023-03-29 01:58:02",
"serial_number": "719770426243590812378787092632593850366518596520",
"subject": "Elastic Certificate Tool Autogenerated CA",
"version": "Version.v3"
}
```

DanOPT:
Converted to draft. Need to have to do a little bit more research.

DanOPT:
Yes, unit tests are probably a good idea. I will implement them.

PR #136, opened 2023-03-29
Allow to set node types for elasticsearch

### LINKED ISSUE #66: Don't use all hosts as potential master hosts
Right now we assume every host is a potential master and use the whole list of hosts for cluster building. If the cluster is expanded in a later stage every new node will be used as master, too.

We need the following:
* Option to set `node.master` to decide which nodes are potential masters and which are not
* Define the list of initial masters during initial setup, write the list into a (fact?)file on the CA host and stick to that in an already deployed cluster
* Some extra logic to check for violations of best practices (odd number of potential masters, not using master only nodes for communication etc)

### LINKED ISSUE #137: Remove initial master node after the first cluster initialize
We should remove the **cluster.initial_master_nodes** after the first cluster initialize otherwise the node will lose data after the first restart. With if condition in the template elasticsearch.yml should we evaluate the cluster status, if it is already built, then do not set **cluster.initial_master_nodes** in the elasticsearch.yml file. 

The roll out of the template should occur two times. One time before the first start of the cluster with notify on the task, because it will be the major config roll out. The first time after the Cluster initialize without notify on the task to not restart all cluster nodes at once. In the future, a restart will take place for any reason and Elasticsearch will read his configuration again.

widhalmt:
fixes #66
fixes #137 

You can now set a new variable to set all the node types. If you don't set it, it will stay at the default of "every node has every role".

Since having an odd count of master nodes is important, we built in a check that will assure you have the right count.


widhalmt:
I changed this to draft because something in the calculation of master nodes isn't right. I tried different approaches with temporary files and other humble makeshift solutions. @afeefghannam89 , @Saeid-Abadi and @Donien showed me a nicer approach. Now it only needs to work. ;-)

Note: The current code *shouldn't* go through the tests. I pushed a draft where the check should actually fail due to a wrong count of master nodes.

widhalmt:
The new approach with setting some variables and creating groups is all but "sophisticated". But it's simple enough that it should work reliably and even more, it has some nice sideeffects that we might benefit from later. Having groups of master-nodes and data-nodes could be beneficial.

Of course I could just have the user set the variables we're setting. But I like the idea of using a list of roles to configure. We don't know whether there will come new ones or what combinations users might use. Then just setting variables we need for easier handling is a bit cumbersome, but hey, it works.

widhalmt:
Thank you so much, @afeefghannam89 for finding https://www.elastic.co/guide/en/elasticsearch/reference/current/bootstrap-checks.html#single-node-discovery This should be it!

afeefghannam89 (approved):

PR #143, opened 2023-04-02
Remove redundant task

### LINKED ISSUE #139: Remove redundante waiting tasks
In `tasks/elasticsearch-security.yml` we have several versions of tasks waiting for Elasticsearch to start.

I guess the following is redundant and can be removed.

```
- name: Check for open port tcp {{ elastic_elasticsearch_http_port }}
  wait_for:
    port: "{{ elastic_elasticsearch_http_port }}"
  tags:
    - certificates
    - renew_ca
    - renew_es_cert
```

I guess, when we remove it, we need to add the tags to the other waiting tasks. @afeefghannam89 you put a lot of effort in those tags, could you plese make sure we don't break anything and remove this task (if I'm right)

afeefghannam89:
Remove a redundant task.
Fix #139

### DIFF
--- a/roles/elasticsearch/tasks/elasticsearch-security.yml
+++ b/roles/elasticsearch/tasks/elasticsearch-security.yml
@@ -604,14 +604,6 @@
   retries: 5
   delay: 10
 
-- name: Check for open port tcp {{ elastic_elasticsearch_http_port }}
-  wait_for:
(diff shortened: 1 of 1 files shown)


widhalmt:
Thanks. I was afraid, that we need to make sure that some tags are present in other tasks as well. But you make it look simple. :-)

widhalmt (approved):

PR #146, opened 2023-04-02
Feature/remove openssl 138

### LINKED ISSUE #138: Remove installation of openssl
It's not best practices to install a package like `openssl` in a collection about Elastic Stack.

Remove the installation and list it as a dependency.

This means, we need to make sure, we have `openssl` in every Molecule check.

afeefghannam89:
Remove the installation of OpenSSL from Elasticsearch role.
Fix #138 
Rework prapare.yml for all  molecule workflows and remove unneeded task to speed up the pipelines little bit.


afeefghannam89:
**Check for cluster status with elastic password** always fails with **full_satck** scenarios for Logstash, Kibana and Beats. This happens in the idempotent check phase, while the cluster status is yellow. I raised the retries to 20, but it does not help. For more information see the **debug** and **fix pipelines**  actions.

On the other hand, this task works fine for elasticsearch pipelines
 Do you have a solution @widhalmt ?

molecule/kibana_default/prepare.yml
```diff
@@ -9,8 +9,7 @@
...
           - procps
           - gpg-agent
+          - git
+          - iproute2
         update_cache: yes
-      when: ansible_os_family == "Debian"
-    - name: Install git
-      package:
-        name: git
+
```
widhalmt, line 15:
```suggestion
```

Extra blank line

roles/elasticsearch/tasks/elasticsearch-security.yml
```diff
@@ -597,13 +597,17 @@
     cut -d\" -f4
   register: es_cluster_status
   changed_when: false
-  no_log: true
-  ignore_errors: "{{ ansible_check_mode }}"
+  # no_log: true
```
widhalmt, line 600: Is this still part of testing? Or do you want to keep it for later to have a reference?

widhalmt (changes requested):
I just marked a bit of lint. Regarding the cluster not building, I need to have a deeper look into the code. I'll come back for that, too.

roles/elasticsearch/tasks/elasticsearch-security.yml
afeefghannam89: > Unfortunately the failing checks are hidden when the linter doesn't run through. I don't know why the cluster stays yellow. I just can assume that there's been a minor problem with the `prepare.yml` playbooks. Maybe the nodes can't connect and every node only checks it's local instance?

The same packages which have been installed before will be installed in this scenario. I do delete only RHEL tasks, because we use Ubuntu and compact the installation Tasks to one.

In the debug action you will see the curl call, which is to the localhost not to the remote host.

I can redo the changes and test.

widhalmt: Hm, but we run tests on Rocky, too. So we need to test both. As it seems, they run through, though. So that shouldn't be the problem.

afeefghannam89:
@widhalmt I have created a new test branch from the actual main branch.  I have removed one empty line in every role and pushed to let the pipelines run for all roles. So the prepare.yml and the complete collection code is unchanged and actual. Anyway the same pipelines for logstash and beats fail with the same reason. That means the changes in this PR are not reason why the pipelines fails, see https://github.com/NETWAYS/ansible-collection-elasticstack/actions/runs/4616563045/jobs/8161826518#step:6:1129

and https://github.com/NETWAYS/ansible-collection-elasticstack/actions/runs/4616563041/jobs/8161756609#step:6:1122

widhalmt (approved):
Nice. Thank you.

PR #153, opened 2023-05-03
add logpath config, change order for path creation

psteinmetzcid:
We need to configure the path for elasicsearch logs. 

We change the folder creation order. First folder creation than start services within elasticsearch-security.yml.
The data directory must exist while first start the elasticsearch service, otherwise the start will fail.


DanOPT:
Thanks for your contribution @psteinmetzcid 

If nobody obligates I will merge your PR soon.

widhalmt (approved):
Nice! Thanks for your contribution. I'll merge it. Welcome to the list of contributers!

PR #154, opened 2023-05-03
Fixed workflow for plugins

DanOPT:
The full-stack workflow for plugins didn't work, since it has no complete molecule scenario and is only executed on the runner. I removed the plugin's scenario from it. https://github.com/NETWAYS/ansible-collection-elasticstack/actions/runs/4862046856


widhalmt (approved):
Looks legit. Should we open an issue to have the plugins scenario run for full stack tests, too? Could help to check before we merge anything that could break the stack.

PR #156, opened 2023-05-12
Connect to Elasticsearch API via IP address

### LINKED ISSUE #155: Creating Logstash users uses hostname and fails if no DNS is available
We used the collection in a test environment and @dnssmr discovered that Logstash can't connect to Elasticsearch for user creation when no DNS resolution is available.

widhalmt:
For environments where we can't use DNS resolution, we should connect to Elasticsearch API via IP address instead of hostname.

fixes #155


dnssmr:
It works, but i think i cant approve because molecule idempotence failed?

dnssmr (approved):
Works

PR #158, opened 2023-05-15
Feature/cluster pipelines test

afeefghannam89:
When we check the Elasticsearch cluster health, we need more than one instance.


afeefghannam89:
The Cluster checks problem will be fixed in #122 so I will delete this PR.

PR #160, opened 2023-05-16
Rename variables to keep naming scheme

### LINKED ISSUE #159: Variable naming conventions
The newest version of `ansible-lint` says, variables within roles should use the role name as a prefix.

For the most part we already stick to that rule. But we have a bunch of variables that are global to the whole collection. They are mostly used for interaction between the roles in the collection. These are prefixed with `elastic_`.

How should we deal with this new rule?

* Just put an exception into the `ansible-lint` configuration?
* Rename the variables so that they are local to every role? If we need interaction we need a lot of `set_fact` for synchronization.
* Come up with a completely different solution?
* Find a way to have "global variables" in collections? A short search didn't reveal any option to do that.

widhalmt:
fixes #159

### DIFF
--- a/docs/role-kibana.md
+++ b/docs/role-kibana.md
@@ -29,7 +29,7 @@ These variables are identical over all our elastic related roles, hence the diff
 * *elastic_kibana_port*: Port Kibana webinterface is listening on (default: `5601`)
 * *elasticsearch_ca*: Set to the inventory hostname of the host that should house the CA for certificates for inter-node communication. (default: First node in the `elasticsearch` host group)
 * *elastic_ca_dir*: Directory where on the Elasticsearch CA host certificates are stored. This is only useful in connection with out other Elastic Stack related roles. (default: `/opt/es-ca`)
-* *elastic_ca_pass*: Password for Elasticsearch CA (default: `PleaseChangeMe`)
+* *elasticstack_ca_pass*: Password for Elasticsearch CA (default: `PleaseChangeMe`)
 * *elastic_initial_passwords*: Path to file with initical elasticsearch passwords (default: `/usr/share/elasticsearch/initial_passwords`)
 * *elastic_release*: Major release version of Elastic stack to configure. (default: `7`)
 * *elastic_variant*: Variant of the stack to install. Valid values: `elastic` or `oss`. (default: `elastic`)
--- a/docs/role-kibana.md
+++ b/docs/role-kibana.md
@@ -20,7 +20,7 @@ Role Variables
 
 These variables are identical over all our elastic related roles, hence the different naming scheme.
 
-* *elasticstack_stack_full_stack*: Use `ansible-role-elasticsearch` as well (default: `false`)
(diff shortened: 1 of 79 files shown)


widhalmt:
I made this a draft because there are still some variables missing in the renaming. We'll discuss how to proceed with them in #159

widhalmt:
We opened a discussion around that issue upstream: https://github.com/ansible/ansible-lint/discussions/3451

To move fast, we now put this on the warn_list and come back to it, when
the discussion is resolved. Hopefully to remove the rule completly from
`ansible-lint.yml`.

dgoetz (changes requested):
If the collection is accepted as prefix in the upstream discussion everything related to the CA shared by the complete stack should be prefixed `elasticstack_`, this goes for unchanged variables like `elastic_ca_pass` and some changed in the PR like `elastic_ca_will_expire_soon`/`elasticsearch_ca_will_expire_soon`. So I would keep thise unchanged until then.

widhalmt:
@dgoetz I hope, I now caught all of the missing variables. Honestly, I feel a slight hint of panic when I think of merging this PR and all the others that are still open. 😱

widhalmt:
Some of the checks are failing because of timeouts. As long as all `lint` checks and the required checks work, we should be safe.

docs/role-kibana.md
```diff
@@ -16,23 +16,23 @@ Role Variables
...
 * *kibana_tls_key*: Path to the key Kibana should use when communicating with clients (default: `/etc/kibana/certs/key.pem`)
 
-* *kibana_security*: Activate TLS and authentication when connecting to Elasticsearch. **Note**: Only works when `elastic_stack_full_stack` is enabled. (default: `true`)
+* *kibana_security*: Activate TLS and authentication when connecting to Elasticsearch. **Note**: Only works when `elasticstack_stack_full_stack` is enabled. (default: `true`)
 
 These variables are identical over all our elastic related roles, hence the different naming scheme.
 
-* *elastic_stack_full_stack*: Use `ansible-role-elasticsearch` as well (default: `false`)
-* *elastic_elasticsearch_http_port*: Port of Elasticsearch http (Default: `9200`)
+* *elasticstack_stack_full_stack*: Use `ansible-role-elasticsearch` as well (default: `false`)
```
afeefghannam89, line 23: How about remove the first stack word?
```suggestion
* *elasticstack_full_stack*: Use `ansible-role-elasticsearch` as well (default: `false`)
```

afeefghannam89:
@widhalmt woo It is a big type work ✍️ Thanks I see the PR is ok. The only suggestion is the above, there are three stack words in the variable name. You can take it or remove it.

afeefghannam89 (approved):

dgoetz (approved):
Naming looks consistent now within the collection and with the guideline (including our proposed addition). It creates some clunky names but that seems unavoidable with the goal in mind. Nitpicking could be done in form of complaining some not being in alphabetical or logical order, but I am ok with this.

widhalmt:
Some of them aren't in alphabetical order on purpose. I tried to group them logically. :-)

PR #164, opened 2023-05-17
Fixed wrong input and output file

dnssmr:
The file /etc/logstash/conf.d/ansible-input/input.conf contained an OUTPUT, while /etc/logstash/conf.d/ansible-input/output.conf contained an INPUT. Fixed that


widhalmt:
Thank you! (Don't mind the failing linters - we need #160 to be merged to fix this issue. It's basically a newly introduced linter role for `ansible-lint` *sigh*.

Issue #157, opened 2023-05-12
gpg package is missing as requirement for debian 11 (and possibly other deb based distros)


dnssmr:
![image](https://github.com/NETWAYS/ansible-collection-elasticstack/assets/118279543/2c91af1f-e85a-4cd8-a274-70b7d075a40d)

gpg is not installed on Debian, and its needed to include the elastic repositories

PR #166, opened 2023-05-22
fixing gpg dependency

dnssmr:
Fixes Issue #157

### DIFF
--- a/roles/repos/tasks/debian.yml
+++ b/roles/repos/tasks/debian.yml
@@ -1,4 +1,9 @@
 ---
+- name: Ensure gpg exists, for signing keys
+  ansible.builtin.apt:
+    name: gpg
+    state: present
+
 - name: Ensure Elastic Stack key is available (Debian)
   ansible.builtin.apt_key:
     url: https://artifacts.elastic.co/GPG-KEY-elasticsearch
--- a/roles/repos/tasks/main.yml
+++ b/roles/repos/tasks/main.yml
@@ -7,11 +7,6 @@
     - elasticstack_release | int > 7
     - elasticstack_variant == "oss"
 
-- name: Ensure gpg exists, for signing keys
-  ansible.builtin.package:
-    name: gpg
-    state: present
-
 - name: Import RedHat tasks
   import_tasks: redhat.yml
   when: ansible_os_family == 'RedHat'
--- a/roles/repos/tasks/main.yml
+++ b/roles/repos/tasks/main.yml
@@ -7,6 +7,11 @@
     - elasticstack_release | int > 7
     - elasticstack_variant == "oss"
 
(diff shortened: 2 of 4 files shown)


dnssmr:
i put the task into the repo role in tasks/main.yml with the package module, since gpg is needed on all OS types

widhalmt (changes requested):
Hm... on rpm based systems the package is called `gnupg2` . We just didn't install it because it's part of the base install. *if* we install it, we should care for both variants, though.

Since I started the roles that moved into this collection I changed my mind, I have to admit. In the meantime I'm not conviced that installing all dependencies with the role is good practice. As I see in many other roles from other people, packages that are not a direct dependency of installed software should be listed in `README.md` as dependency and not be touched by the role itself.

Of course, we need to make sure they are still installed via `prepare.yml` in Molecule checks.

Maybe we should use this discussion as an example on how we want to proceed with the roles and collections in our namespace? I'm ok with both ways, I just want a "default way" of doing it.

widhalmt:
The "nuke 'em from orbit" variant would be to install the packages but allow for deactivating this task via variable. Should make everybody happy. Ist just a significant bigger blob of code.

widhalmt:
@dgoetz had another idea: Manage all packages without alternatives inside the role. Especially if they are part of default repositories. Make sure there's no version pin. Everything else (optional packages, those that need extra repos etc.) added as task but disableable (is that a word?) via variable.

dnssmr:
Just vomiting thoughts here:

I think we should add dependent packages in the role, as long as we know that they  could be missing (like gpg), because maintaining ALL dependencies is nearly impossible and a lot of work, and if they are in the base repos. If they need extra repos or be downloaded or something, they should be diableable (it is a word now). 

BUT: Imagine this scenario: We need a package that has its own repo and stuff (like something from hashicorp or something) and if we dont have the package, installation of other things fails and we cannot provide the goaöl of the role. Will we say in the readme "yeah you can disable it, but everything will fail"? I thinkwe should say something like "package x will be provided from source y, if you want to provide it from another source, you can set install_package_x to false. But make sure you have it installed before it runs"

widhalmt:
I was just thinking of the, hopefully mostly hypthetical, situation where users want to install "their own(tm)" version of `gpg`. I've seen strange things in my consulting days... But to be honest, I guess, if someone *really* wants to do such a thing, they can always open an issue.

So +1 from my side for your and your and Dirks approach: Install everything we need if it's mandatory for the role to run and mandatory. And install everything else but make the task disableable (🎉 ).

widhalmt (approved):
Thanks a lot for the changes.

I checked again, the package name on RedHat is `gnupg2` currently but the package provides `gnupg`. So we even cover name changes etc. Great!

Approved.

PR #168, opened 2023-05-23
Replace rocky with centos due to test errors

widhalmt:
It's just a test if we are facing a problem in GitHub or in the `rockylinux8` containers.


widhalmt:
It can be seen as verified that the problem is in fact not the usage of Rocky Linux containers. So I'm closing the PR

PR #171, opened 2023-05-30
Reduce pipelines

### LINKED ISSUE #170: Rework GitHub pipelines
Some pipelines like elasticstack_default run many times without changes in between before merge PR in main  branch. This takes too long time.

afeefghannam89:
* Fix #170  remove repetitive pipelines
* Make full stack pipelines run manually before release. 

We can run this very big pipeline before release manually on a specific branch e.g. main. When we get error we can fix it in another branch and then merge in main. The benefit from this:
* We should not wait long time before every PR merge. We wait only before the release.
* It is easier to debug. We should not wait on `test_role_pr` workflow then on `test_full_stack` to see if our fix work or not, when we have problem in the `test_full_stack`. The waiting time is about 3.5 hours. Now we can comment out all container, where our code worked, then test fixes on not working containers and run the CI/CD manually until we get every thing fixed.

### DIFF
--- a/.github/workflows/test_full_stack.yml
+++ b/.github/workflows/test_full_stack.yml
@@ -1,12 +1,17 @@
 ---
-name: Test ElasticStack
+name: Test ElasticStack before release
 on:
-  push:
-    tags:
-      - '*'
-    branches:
-      - main
-  merge_group:
+  workflow_dispatch:
+    inputs:
+      logLevel:
+        description: 'Log level'
+        required: true
+        default: 'warning'
+        type: choice
+        options:
+          - info
+          - warning
+          - debug
 
 jobs:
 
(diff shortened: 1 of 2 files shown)


danopt (approved):
Looks good!

widhalmt:
Good idea! But this means, we need to change branch protection. If we don't have the full stack tests running before merging into `main` we can't have these checks as a prerequisite for merging. Like this PR shows, you can't merge because the checks aren't running. But not running is exactly what this branch is about.

On the other hand, that means, we need to decide which checks are `required` for every merge and which are just informational.

afeefghannam89:
> Good idea! But this means, we need to change branch protection. If we don't have the full stack tests running before merging into `main` we can't have these checks as a prerequisite for merging. Like this PR shows, you can't merge because the checks aren't running. But not running is exactly what this branch is about.
> 
> On the other hand, that means, we need to decide which checks are `required` for every merge and which are just informational.

@widhalmt you are right big part of the implementation is done in another PR, I would test some scenarios here, to see what is better for us, then I will share you the results :)

PR #174, opened 2023-06-06
fix missing default var

### LINKED ISSUE #173: Missing variable elasticsearch_check_calculation
The variable `elasticsearch_check_calculation` has no default.   
So if not set, the role will fail.

Refers to #148

tbauriedel:
adds missing default var to elasticsearch role.

fixes #173


dnssmr (approved):

afeefghannam89:
Thanks @tbauriedel

PR #176, opened 2023-06-13
Update README.md

### LINKED ISSUE #175: Enhancement : Add Variable for Version(s) which should be installed
Currently the collection is always installing the latest versions of the elastic stack
But sometimes this is not wanted, for example when adding a new node to an existing ES Cluster without updating the whole cluster. 

It would be very useful to have the versions configurable via  variables

martialblog:
Tried to update the README to make some things clearer. Also fixed the formatting of the Versioning section to be more readable. 

Fixes #175

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,9 @@
 
 ![Test ElasticStack](https://github.com/NETWAYS/ansible-collection-elasticstack/actions/workflows/test_full_stack.yml/badge.svg)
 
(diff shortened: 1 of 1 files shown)


martialblog:
@widhalmt @DanOPT there might be an issue with the GitHub Actions Config. I assumed the `doc/` branch name would not have triggered the molecule CI

dnssmr (approved):

afeefghannam89:
> @widhalmt @DanOPT there might be an issue with the GitHub Actions Config. I assumed the `doc/` branch name would not have triggered the molecule CI

@martialblog you are right, I will open an issue for that.

PR #179, opened 2023-06-15
Refresh apt cache before installing packages

### LINKED ISSUE #167: Update Debian apt cache when packages are not installed
I'm not as used to Debian as I'm used to RedHat, so take this with a grain of salt: Should we update `apt` cache before installing the packages? There are similar tasks in out Molecule checks but not in the actual roles.

In https://github.com/geerlingguy/ansible-role-mysql @geerlingguy has a task that checks whether the package is already installed. And if not, then her refreshes the `apt` cache. Should speed up the process significantly.

widhalmt:
Run an apt refresh at the start of each role. We can use grace time so we don't neet to refresh it every time and safe some time.

fixes #167


widhalmt:
It seems like we need to merge #182 before continuing with this.

dnssmr:
Didn't we fix the tests? Can we merge main?

widhalmt:
I thought the tests were fixed. :-( As much as I see, tests are still timing out. I don't want to remove the thorough tests we built so far. So I made tests "non mandatory" and check them manually. While this is ultra cumbersome I guess it's the best solution we have so far. Even more so as @martialblog said, we'll have our own GitHub runners soon. And then we'll definitely want our checks back.

So please see the checks "more like guidelines, anyway" and approve if the code makes sense. Even when some checks time out.

roles/beats/tasks/main.yml
```diff
@@ -6,6 +6,12 @@
     - '{{ ansible_os_family }}_{{ ansible_distribution_major_version }}.yml'
     - '{{ ansible_os_family }}.yml'
 
+- name: Update apt cache.
+  apt:
```
dnssmr, line 10: >:(

roles/elasticsearch/tasks/main.yml
```diff
@@ -1,4 +1,11 @@
 ---
+
+- name: Update apt cache.
+  apt:
```
dnssmr, line 4: Please use fqcn

roles/kibana/tasks/main.yml
```diff
@@ -1,5 +1,11 @@
 ---
 
+- name: Update apt cache.
+  apt:
```
dnssmr, line 4: ! >:(

roles/logstash/tasks/main.yml
```diff
@@ -1,5 +1,11 @@
 ---
 
+- name: Update apt cache.
+  apt:
```
dnssmr, line 4: Here as well

dnssmr (changes requested):
I am a linter

widhalmt:
Thanks so much @dnssmr I guess, I'll take a while before I adapted my old brain to this new style of coding. 🤦🏽

widhalmt:
I hope, I caught all of them, now.

widhalmt:
I converted this to da draft because the tests found some error that's not just due to problems in GitHub runners. I'll investigate locally and then come back. On the other hand... that's out of the scope of this PR, so I'll disable Draft status. 🤦🏽

dnssmr (approved):

widhalmt:
We need #279 merged for this to work and be merged.

PR #182, opened 2023-06-16
Replace `molecule-docker` with `molecule-plugins[docker]`

widhalmt:
Moving to newer versions of molecule, this seems to be neccessary. Refer to e.g. https://dailystuff.nl/blog/2023/switch-to-molecule-plugins


afeefghannam89 (approved):
Thanks you very much :)

PR #187, opened 2023-06-19
set users password as hash

### LINKED ISSUE #126: Support password hash for users
We should enable the usage of hashed user password. By the way, we should enforce the setting of password long to => 6 characters. This what the documentation point out.

afeefghannam89:
Fix #126 Enable hashed password for users


widhalmt (approved):
That's a nice addition! I like it.

PR #188, opened 2023-06-19
Modify the certificate validity period

### LINKED ISSUE #127: Enable the modification of certificate validity
When the collection certificate or ca produce, the validity will be set to default duration of 1095 days. It would be cool to let user modify this

afeefghannam89:
Fix #127 Enable the adjustment of certificate validation


widhalmt:
I'm not totally sure. But wouldn't omitting the value just remove the value? And we'd end up with `--days` without a value?

afeefghannam89:
> I'm not totally sure. But wouldn't omitting the value just remove the value? And we'd end up with `--days` without a value?

You are right, thanks for the hint:) I have updated the tasks and the docs. I hope the PR is now ok.

Issue #172, opened 2023-05-31
Merge or squeeze logstash CI/CD pipelines


afeefghannam89:
I think we can merge `logstash_defualt` and `logstash_pipeline` to one CI/CD pipeline. I think that the installation of logstash as standalone and testing his pipelines in one step will reduce the CI/CD running time and achieve two goals at once.

I am not sure if we can merge `logstash_run_logstash` with another CI/CD pipelines.

@widhalmt what do you think?

widhalmt:
Good idea. I wanted to have both separated to test the role with as many defaults as possible. To show that the collection has a very low entry threshold. But since we have to set some variables nonetheless we can just merge one or both, too.

`logstash_run_logstash` was mostly introduced to check whether Logstash would start or not. So that we could have other pipelines without starting Logstash thus being faster. But I guess, that also lost its use now we change so much.

So yes, from my side, go ahead and merge them.

PR #190, opened 2023-06-20
Enhancement pipelines

### LINKED ISSUE #180: CI/CD pull request test run while updating documentation
pull request CD/CI checks run when we update README.md or docs. We should restrict this to the concerned files/directories.

### LINKED ISSUE #189: Fix logstash pipeline
Logstash pipeline fails, because of the usage of the latest python version with molecule test.

afeefghannam89:
* Restrict the pull request workflow, fix #180 
* Use python 3.8 in Logstash pipeline, fix #189 
*  Only run pr workflow while opening a pull request, but not in merge queue #193 
* Merge or squeeze logstash CI/CD pipelines #172
* Fix full stack workflow on all containers #194


widhalmt:
I see parallel checks. Won't that lead to collisions where different instances could run in the same container and so break a run?

afeefghannam89:
> I see parallel checks. Won't that lead to collisions where different instances could run in the same container and so break a run?

N, I do not think so. Every instance will run separately. Did you see under the GitHub action history?

widhalmt:
No, I didn't see it now. It just occured to me, that in some tests I had the same test breaking on different stages for no apparent reason. On closer analysis I found out that another Ansible run was changing things the current test just made. Thus it became apparent that several tests were using the same container.

But I have to admit, this happened long back and might not be an issue at all in current setups.

The easy solution, just renaming the containers in `molecule.yml` doesn't work if you run the same test with different variables from a matrix.

widhalmt (approved):
This will greatly improve our testing. Thank you a lot!

PR #197, opened 2023-06-23
Add notice about sponsor

widhalmt:
We have the first sponsor who aggreed to be mentioned with their name! Here's a way to show them our appreciation.


tbauriedel (approved):
Looks good to me

PR #199, opened 2023-06-30
Fix/lint workflow 192

### LINKED ISSUE #192: Take out lint jobs from all workflows to a separated one
It will be good to take out all lint jobs from all workflows files and build a separated linter workflow. Other corresponding workflows will only run if linter workflows ended successfully. What we achieve with this?

01- Compact code
02- Now, many pipelines/workflows will run if a change happens on `.config/ansible-lint.yml` or `.yamllint`. Only linter pipeline should be run in this case.

Donien:
Fix #192.

This PR introduces a reusable workflow. This general workflow handles the job necessary to run both `ansible-lint` and `yamllint` on **either** all roles or one specific role.  

All roles are linted when:
- linting rules are changed
- the linting workflow is changed

Specific roles are linted when:
- a role specific workflow is triggered (linting as first job)
- the `full_stack` workflow is triggered (linting as first job)

The workflow for the specific role now uses a call to the general linting workflow, while providing it with its own role name.

This approach does not completely remove linting from single workflows, but it makes those linting jobs much more readable and reduces redundant code, while also allowing for a bit more flexibility.

### DIFF
--- a/.github/workflows/test_full_stack.yml
+++ b/.github/workflows/test_full_stack.yml
@@ -17,30 +17,8 @@ on:
 
 jobs:
 
-  lint_full:
-    runs-on: ubuntu-latest
-    steps:
-      - name: Check out the codebase.
-        uses: actions/checkout@v3
-
-      - name: Set up Python 3.
-        uses: actions/setup-python@v4
-        with:
-          python-version: '3.x'
-
-      - name: Install test dependencies.
-        run: |
-          python3 -m pip install --upgrade pip
-          python3 -m pip install -r requirements-test.txt
-
-      - name: Lint code.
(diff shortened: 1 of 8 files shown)


widhalmt:
Nice! That seems like it could really help a lot with development.

widhalmt:
We now have merged some major changes to the testing configuration. Could you please resolve the conflicts and have another look?

Donien:
I've looked through the workflows again and made some changes.  
This still needs a review. But as of today it should be compatible with the current main branch.

widhalmt (approved):
Thanks again. This really will help with easier maintenance.

PR #200, opened 2023-06-30
Change used module from 'command' to 'shell' to allow shell redirect

### LINKED ISSUE #198: Suspected bug: "Enable Ingest Pipelines" task not working due to impossible redirect
Within **roles/beats/tasks/filebeat.yml** there is the following task:

```yaml
    - name: Enable Ingest Pipelines
      command: >
        /usr/bin/filebeat setup --pipelines &&
        /usr/bin/filebeat version > /etc/filebeat/{{ item }}_pipeline_created
      args:
        creates: "/etc/filebeat/{{ item }}_pipeline_created"
      with_items: "{{ beats_filebeat_modules }}"
      notify:
        - Restart Filebeat
      changed_when: false
```

It uses the `command` module. Within its shell command a redirect is used (`>`).
This leads to a problem: The redirect into the file does not work using the `command` module since no actual shell is used.  
It does however work using the `shell` module. We can even keep the `creates` argument, making this a pretty simple fix.

Donien:
Fix #198

This PR adjusts a task that used the `command` module with a command that wants to redirect output into a file.  
This, however, is not possible since no actual shell is used, thus making that redirect impossible. The Ingest Pipelines were never created.  
Changed the `command` module to the `shell` module.

### DIFF
--- a/roles/beats/tasks/filebeat.yml
+++ b/roles/beats/tasks/filebeat.yml
@@ -87,7 +87,7 @@
         - elasticstack_release | int > 7
 
     - name: Enable Ingest Pipelines
-      command: >
+      shell: >
         /usr/bin/filebeat setup --pipelines &&
         /usr/bin/filebeat version > /etc/filebeat/{{ item }}_pipeline_created
       args:


widhalmt:
Great find! Thanks. Let us know, when you're done so we can review it. Don't let yourself be let down by failing checks. That's more due to GitHub problems then the code.

afeefghannam89 (approved):

PR #201, opened 2023-06-30
Fix some 'changed_when: false' to reflect actual changes

### LINKED ISSUE #140: Rethinking the changed_when: false
I think, in some tasks checnged_when false does not serve any goal, we should check them again

Donien:
Fix #140 

Needs the fix of #198 (potentially fixed by #200)

This PR changes a few 'changed_when: false' statements to 'changed_when: true' to better reflect actual changes being made.  
Those tasks already check if e.g. passwords present differ from the ones we request. Since those specific tasks are only run if we see the need for a change, it should be safe to consider it a change if they do run.


widhalmt:
Setting `changed_when` to `true` will break idempotence tests, won't it? Or am I totally mislead now?

Donien:
It *would* destroy idempotency. But as I tried to lay out we already do check whether a change is needed and only then run the appropriate tasks if needed.  
So we can be sure that we will actually do changes now.  
Thus we can also set changed_when to true, since those tasks run **only if thoses changes would be made**. Otherwise they are skipped.  
Setting changed_when to true is only done so ansible-lint does not complain.

If the task is not run, changed_when won't break idempotency. If it is run, we expect (and get) a change anyway.

widhalmt:
Damn, totally missed that. I'm so sorry. That's definitely the best of the possible solutions to this problem.

widhalmt (approved):
Your explanation makes sense. So: Approved.

PR #211, opened 2023-08-03
Allow to set DN of CA

### LINKED ISSUE #210: Allow naming of Elasticsearch CA
There's a simple option to name the Elasticsearch CA. A customer requested to have the option to change the default name. We'll put that in if the default set to the current default so it won't change anything on already running systems if you ignore the setting.

ref/NC/791092

widhalmt:
This adds the `--ca-dn` parameter to the creation of the Elastic Stack CA. The default is the same as if you would just omit the parameter.

For more details refer to https://www.elastic.co/guide/en/elasticsearch/reference/8.9/certutil.html#certutil-parameters

fixes #210

ref/NC/791092


DanOPT (approved):
Looks good! [👍](https://emojipedia.org/thumbs-up/)

PR #213, opened 2023-08-04
Install cryptography Python library

### LINKED ISSUE #212: Automatically install Python cryptography module
Elasticsearch needs the `cryptography` Python module to create a CA and certificates. Usually we leave it to the user to choose which installation method they prefer.

Since today we got another user stumbling over this missing dependency I want to implement the installation now automatically.

widhalmt:
We had this as a requirement. But it makes sense to have the role install the library without any further interaction with the user.

fixes #212


Saeid-Abadi (approved):

DanOPT:
Hmm, I checked it on different OS. As far as I can see, all newer OS use python3 and the package is available under python3-cryptography. The only one I know where python-cryptography >= 2.5 is not available within the system package manager is CentOS 7. There we will need to install it with pip. 

That should be convention after PEP 394 and most distributions following this recommendation except for Windows. (https://stackoverflow.com/questions/62214293/is-python3-always-installed-with-python-3)

I double checked it on these OS:
Rockylinux 8
Python 3.6.8 (default, Jun 22 2023, 05:10:43)
python3-cryptography

Debian 11
Python 3.9.2 (default, Feb 28 2021, 17:03:44) 
python3-cryptography

Ubuntu 20.04
Python 3.8.10 (default, May 26 2023, 14:05:08) 
python3-cryptography

Ubuntu 22.04
Python 3.10.12 (main, Jun 11 2023, 05:26:28) [GCC 11.4.0] on linux
python3-cryptography

For older systems the user can change the package to python-cryptography instead of python3-cryptography.

widhalmt:
I guess, you're right. I'm just living in the past.... I go for `python3-cryptography` then.

widhalmt:
I'll wait for #179 and #243 to be merged before resolving these conflicts. This should give is working tests. Finally.

widhalmt:
Looks like repositories of Rocky Linux are facing some issues right now. I'll try later and rerun the failing tests.

afeefghannam89 (approved):
Looks ok to me :)

Issue #217, opened 2023-08-08
Typo in Variable present in whole code for ealsticsearch_http_publish_host


xtruthx:
https://github.com/NETWAYS/ansible-collection-elasticstack/blob/22c2a9dd51b6f5a58dfa3c985245b2fd70deb368/docs/role-elasticsearch.md?plain=1#L34

*ealsticsearch_http_publish_host* Typo is present in the whole code like tasks and templates.

PR #220, opened 2023-08-09
make elasticsearch network.host configurable

### LINKED ISSUE #219: Posibility network.host as variable
Setting `network.host` with hardcoded `"_local_", "_site"` is not very smart in enviroments with multiple publich interfaces e.g if you have something like IDRAC interfaces or you want to separate network usage.

So the the template should be adopted to be able to declare it as a variable.

xtruthx:
This PR fixes  #219


widhalmt (approved):
Thanks for the PR!

I'm just wondering, should we use a single string for the variable or should we make the variable an array since the option in the file is an array, too?

I'm ok with your string variant, especially since using Jinja2 for an array would be much more complicated (like ifnotlast, then "," etc)

So, approved. Thanks, again.

widhalmt (changes requested):
Sorry, I had to change back to "Request changes". Could you please write a line about the new variable in the appropriate file in `docs/`?

xtruthx:
> Sorry, I had to change back to "Request changes". Could you please write a line about the new variable in the appropriate file in `docs/`?

I will avoid doiing duplicate work. This will be done within the PR for the documentation review. Pls accept the PR without additionl docs, anyway if you not set it the "legacy" default will take precedence.

widhalmt (approved):
Ah, ok. That's fine, of course. So - approved.

xtruthx:
the fix of #217 conflicts :/

dnssmr:
Its just a renaming of variables, you should just replace the names and it should work. I can do it quickly if you want to?

xtruthx:
> Its just a renaming of variables, you should just replace the names and it should work. I can do it quickly if you want to?

thx, its already done.

widhalmt (changes requested):
Sorry, I have to revert my approval. I thought a lot about how the collection is proceeding and I'm afraid we have to face the fact that we aren't full time developers andtherefore we can be sure that we run into times where changes will have to wait for a long time. (You had to experience that for yourself). So the only way to not lose focus and have all neccessary steps done right away is to make changes atomic. Therefore I opened an issue to force documentation with every step: https://github.com/NETWAYS/ansible-collection-elasticstack/issues/234 .

So please add documentation for that variable with this PR.

xtruthx:
> Sorry, I have to revert my approval. I thought a lot about how the collection is proceeding and I'm afraid we have to face the fact that we aren't full time developers andtherefore we can be sure that we run into times where changes will have to wait for a long time. (You had to experience that for yourself). So the only way to not lose focus and have all neccessary steps done right away is to make changes atomic. Therefore I opened an issue to force documentation with every step: #234 .
> 
> So please add documentation for that variable with this PR.

I've add a description to the documenation.

Issue #178, opened 2023-06-15
Enhancement: Use FQCN for Tasks


dnssmr:
I know, some people like the FQCN and some hate them, but i'd like to refactor the repository to use the FQCN since Ansible recommends it to avoid conflicting with other collections that may have the same module name.

widhalmt:
Well, I *do* hate them. And even some people who contribute regularly to Ansible hate them. But as you know, I'm always ok with change if we can people make happy. Let's hear what others who contribute to this repository, think.

I'd prefer to ignore FQCN but if only the two of us raise their voice, I'm ready to go with your approach.

PR #224, opened 2023-08-11
Feature/fqcn 178

dnssmr:
Uses FQCN for Ansible Modules. "Foxes" Issue #178


widhalmt (changes requested):
While I still hate this syntax, I see your point. We really should stick to community standards. So, thank you very much for this huge contribution.

The only thing left is that you earned the honor to remove https://github.com/NETWAYS/ansible-collection-elasticstack/blob/main/.config/ansible-lint.yml#L17 with this epic change.

dnssmr:
> While I still hate this syntax, I see your point. We really should stick to community standards. So, thank you very much for this huge contribution.
> 
> The only thing left is that you earned the honor to remove https://github.com/NETWAYS/ansible-collection-elasticstack/blob/main/.config/ansible-lint.yml#L17 with this epic change.

Done! :D

widhalmt (approved):
Perfect! Thank you.

widhalmt:
Maybe that were some new contributions? After removing the exemption for ansible lint, at least we'll see them right away.

PR #230, opened 2023-08-15
Check for license to be set

widhalmt:
I had problems with setting up a single node stack vers. 8.

Here's the included changes:

* Check for license (I had errors showing that no license was yet created when trying to restart)
* Run checks even in single node clusters
* Type on `jvm.options`


widhalmt:
I'm not totally convinced that we really need this change. I had a problem with a non existent license but this might have been just a side effect of the typo in the heap size.

widhalmt:
I guess this would just fix a very rare condition only present in a single testing scenario. So I'm closing it and won't merge the branch.

Issue #42, opened 2023-02-03
Check if certificates work with current passphrase


widhalmt:
When changing a certificates passphrase it will be changed in the configuration using the certificate but not on the certificate itself. We need to implement a check whether the certificate works with the current passphrase or not. If not, we need to regenerate it (maybe keeping a backup of the old one).

Please refer to #41 , #40 and #35 so all the changes will work together.

DanOPT:
Checking passphrases should be possible, but we can't be sure if the file is corrupted or the passphrase is wrong. I guess that wouldn't be a big problem, because in both cases the certificates should be regenerated.

Currently, if the passphrase is wrong or the certificate is corrupted, the cert_info module will fail with a `ValueError` exception. We can catch the exception and return the result of the check without failing the module.

Here is a test case for a wrong passphrase: https://github.com/NETWAYS/ansible-collection-elasticstack/actions/runs/4818065211/jobs/8579587032#step:6:84

@widhalmt Do you see any problems if the module can't difference between invalid PKCS12 data or a wrong passphrase?

widhalmt:
Nope, no problem at all. Please go for it.

DanOPT:
Alright 👍

DanOPT:
- [ ] #238
- [ ] Implement check in tasks and renew certificates, if required because of corrupted data or passphrase change

PR #238, opened 2023-08-21
Feature: Certificate passphrase check

danopt:
#42 
- Introduces a new parameter to cert_info module named `passphrase_check`
- If `passphrase_check` is set to true, the module will only check if current passphrase is working and returns the result of it, but will not return any information about the certificate

### DIFF
--- a/.github/workflows/test_full_stack.yml
+++ b/.github/workflows/test_full_stack.yml
@@ -31,7 +31,6 @@ jobs:
 
     strategy:
       fail-fast: false
-      max-parallel: 2
       matrix:
         distro:
           - rockylinux8
--- a/.github/workflows/test_plugins.yml
+++ b/.github/workflows/test_plugins.yml
@@ -66,7 +66,6 @@ jobs:
 
     strategy:
       fail-fast: false
-      max-parallel: 1
 
     steps:
       - name: Check out code
@@ -111,7 +110,6 @@ jobs:
 
     strategy:
       fail-fast: false
-      max-parallel: 1
       matrix:
         python_version: [ 3.5.10, 3.6.15, 3.7.13, 3.8.16, 3.10.10 ]
 
@@ -151,7 +149,6 @@ jobs:
 
     strategy:
       fail-fast: false
-      max-parallel: 1
       matrix:
         ansible_core_version: [ 2.11.12, 2.12.10, 2.13.8, 2.14.4 ]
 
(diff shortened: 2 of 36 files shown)


plugins/modules/README.md
```diff
@@ -14,7 +14,7 @@
 
 ## `cert_info` module
 
-The netways.elasticstack.cert_info module gathers information about pkcs12 certificates generated by the Elastic stack cert util.
+The netways.elasticstack.cert_info module gathers information about pkcs12 certificates generated by the Elasticstack cert util.
```
martialblog, line 17: I think Elastic writes it like so "Elastic Stack"

tests/unit/plugins/modules/test_cert_info.py
```diff
@@ -131,6 +143,24 @@ def test_module_exit_when_path_and_password_correct(self):
             })
             cert_info.main()
 
+    def test_module_exit_when_password_wrong_with_passphrase_check(self):
+        with self.assertRaises(AnsibleExitJson):
+            set_module_args({
+                'path': 'molecule/plugins/files/es-ca/elastic-stack-ca.p12',
+                'passphrase': 'PleaseChangeMe-Wrong',
+                'passphrase_check': True
```
martialblog, line 151: I might be wrong, shoudn't this be False? 
If not I might be misunderstanding the code. Maybe a comment here would be nice. 

Other question, should there be a test in which passphrase_check: False is tested?

martialblog (commented):
Code looks sound. 

Just have some questions. Otherwise looks very good.

tests/unit/plugins/modules/test_cert_info.py
DanOPT: Added a comment for clarification. passphrase_check is already set to false by default. So all the tests that already existed are testing these cases.

plugins/modules/README.md
DanOPT: Yes, I corrected it.

tests/unit/plugins/modules/test_cert_info.py
DanOPT: Or do you mean, the functionality of passing the parameter should also be tested?

martialblog: I was just wondering if the 'passphrase_check': True was tested. If it's the default, then I think it's alright.

martialblog:
:+1: LGTM

martialblog (dismissed):

widhalmt:
Sorry, I totally missed this one: Please add https://github.com/NETWAYS/ansible-collection-elasticstack/blob/feature/crt_passphrase_check/.github/workflows/test_roles_pr.yml#L33 to your GitHub action. Otherwise GitHub will not start mandatory checks while in the Merge Queue.

widhalmt (dismissed):
Most steps were already approved and tested. So here's my approve for automated tests in the merge queue.

PR #186, opened 2023-06-19
Feature/use uri not curl 123

dnssmr:
Close #123
Close #266 
Close #265 
Close #11

PR #240, opened 2023-08-22
Minor fix for Debian dependencies.

widhalmt:
Sorry for not creating an issue beforehand


widhalmt:
I know @dnssmr has a fix in #186 that would make `curl` installation obsolete. We can happily remove `curl` installation when the PR is merged.

dnssmr (approved):

PR #241, opened 2023-08-22
Allow unsafe shutdown in Logstash

### LINKED ISSUE #239: Logstash restart hangs during tests
Logstash in it's default configuration won't restart when there are messages in the pipeline that can't be written to an output.

There are two possible solution which should both work:

* Use persisted queues for Logstash. Make this optional but use it in every Molecule check just to be sure
* Set a parameter to allow unsafe restarts for Logstash

Personally I'd like to have both implemented.

widhalmt:
Also:
* Rearrange YAML code in pipelines.yml template so it's easier to maintain
* Rename broken variable names
* Activate new feature for molecule scenarios that use Logstash
* Minimalistic docs for new feature

fixes #239


dnssmr (approved):
Looks good! Thank you!

PR #249, opened 2023-08-29
Make logstash_writer user password hashing idempotent

### LINKED ISSUE #244: Make `no_log` disableable
While `no_log` is essential for production environments, it's a major PITA while debugging.

My suggestion: Set all `no_log` options to a global variable called `elasticstack_nolog` which defaults to `true` in every role. Then set it to `false` in every Molecule scenario. And add some documentation telling users to use it while developing.

### LINKED ISSUE #247: Logstash hashing breaks idempotency?
The hashing in https://github.com/NETWAYS/ansible-collection-elasticstack/blob/main/roles/logstash/templates/logstash_writer_user.j2#L3 seems to break idempotency. So I had to deactivate it for Molecule in #243 .

This issue is for investigating if there really is a problem with the hashing. e.g. creating different hashes with repeated runs. Maybe this will also help with solving #186

### LINKED ISSUE #251: Place jvm option before Elasticsearch start
The jvm option should be placed before Elasticsearch start/restart. When the proper jvm options are set **after** the elasticsearch start/restart, the service will fails to start also checking the cluster state with api will also fail when the available resource for the instances are too small.

widhalmt:
Using randomly created hashes for our hash algorithm will break idempotency. Besides, it will lead to the `logstash_writer` user template being recreated with every run. I'm not sure if changing the contents every time can't have any more side effects we can circumvent by just using a fixed hash.

fixes #247
fixes #251
fixes #244

### DIFF
--- a/.config/ansible-lint.yml
+++ b/.config/ansible-lint.yml
@@ -7,7 +7,6 @@ warn_list:
   - key-order[task]  # Ensure specific order of keys in mappings.
   - name[casing]
   - 'risky-shell-pipe'
-  - no-handler # backup of old certificates
   - var-naming[no-role-prefix] # remove when https://github.com/ansible/ansible-lint/discussions/3451 is done
 skip_list:
   - '106'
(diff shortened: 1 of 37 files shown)


widhalmt:
Please see the ongoing discussion about this solution in the linked issue (#247 )

PR #250, opened 2023-08-30
Add code for automatic updates

### LINKED ISSUE #205: Logstash version number structure
We added elasticstack_version to inventory as variable to all Hosts in the Cluster.
We have found that the Version number for Logstash in ubuntu is not the same as for all other products.

The Structure is e.g. '1:8.8.1-1' for Version 8.8.1.

I think it would be good to rebuild the task that compiles the version number for Logstash. This way, elasticstack_version does not have to be maintained multiple times in the playbook.

### LINKED ISSUE #216: No handling for elasticsearch cluster updates (rolling updates)
We installed a Elasticsearch cluster using the collection in Version 8.8.1. After the successful installation we tried a update of the cluster to the (at the moment) latest version 8.9.0. 
It turned out that the collection is installing the new Elasticsearch version but does not restart the services afterwards so the node stays on the old version.
We investigated a bit deeper and it looks like there is no handling for a rolling cluster update at all. There is no serial restart of the services after an update and no cluster health check after a service restart implemented. 

Are there any plans to integrate these cluster update scenarios?

### LINKED ISSUE #236: Read installed version and use as new default
Elastic Stack doesn't like having different versions of services throughout one setup.

What about the following:
* If you don't give a certain version to install, we install the latest one available in the repository. Just like now
* With every Ansible run after the first, we read the installed version (e.g. from the CA host) and use this as the new default

This way we could make sure that new hosts always have the correct version, even when they are set up some time after the initial setup.

### LINKED ISSUE #286: Should we introduce an extra role for collection wide variables?
Like in https://stackoverflow.com/questions/67213727/ansible-collections-collection-level-variable we could build a role called `elasticstack`. This will only hold `defaults/main.yml` and `README.md`. All variables we use in all roles could be set there. So we had them in a central place to change and document. And we could get rid of those pesky linter messages about variable names not matching role names.

widhalmt:
Adding playbook contributed by @xtruthx

I have to admit, I kept this playbook for waaaaay too long. @xtruthx sent it to me a long while back and I always planned to integrate it into the codebase of this collection.

Now I had to face that I took too long and so I put it up publicly. Now I worked it over so that it will fit into the `elasticsearch` role just well.

This PR also introduces a new role where all "global" variables can be found. This will help determining the correct version of a package to install. Some tasks that were repeated in every role get moved into this role called `elasticstack` to streamline execution.

This PR now reads the current version of Elasticsearch installed on the CA host and uses it as new installation target for all further installations. This not only helps with upgrading, it will also make sure that you won't have a version mismatch if you install components after the initial install.

Special thanks to @hryamzik for the workaround to run a task file like with `serial: 1`. Found in https://github.com/ansible/ansible/issues/12170#issuecomment-372263149

Very special thanks to CID GmbH for sponsoring parts of the the development of this PR.

Things that still need to be done:

- [x] Introduce a variable holding the elasticsearch nodename of each node. The check whether the host is back in the cluster relies on it. For now it only works with the hostname, but what if we changed the hostname?
- [x] Create a way to recover from upgrades gone wrong. Especially when the play had to be canceled. Either add documentation on how to recover or even find an automatic way. (e.g. if you turn off shard allocation, then the cluster will not always return to `green` or `yellow` while a node is broken

fixes #216
fixes #286 
fixes #236
fixes #205


widhalmt:
I missed the notes that came with the script (translated by yours truly):

* No version check. It's used to update servers in the project even when no Elasticsearch upgrade is due
* Only for `.deb` so far
* Just "rolling upgrade" method
* Just https and user authentication supported
* No clever error handling
* Reboots server if needed

roles/elasticsearch/tasks/main.yml
```diff
@@ -126,8 +126,10 @@
       }}
 
 - name: Update Elasitcsearch if needed
```
dgoetz, line 128: Can you fix the typo "Elasitcsearch" while on this?

widhalmt: Thanks. Committed the fix locally, will push later.

widhalmt:
The rolling restart via Handler in Elasticsearch doesn't work like the task in the same role. Even though I merely copied and weeded out the code.

So for now we have something that's "resembling it a bit". There's separate runs but every task is run on every node. It doesn't wait for all tasks being finished on one node. I guess this will have to stay as it is for now. And we have to hope that config changes don't appear too often after the initial install. Or we find another way. Anyway, I will add an issue and keep it how it is for now.

roles/elasticsearch/tasks/elasticsearch-rolling-upgrade.yml
```diff
@@ -0,0 +1,204 @@
...
+  when:
+    - elasticsearch_running.status.ActiveState == "active"
+    - groups['elasticsearch'] | length > 1
+  block:
+
+    # Usually we should not need this step. It's only there to recover from broken upgrade plays
+    # Without this step the cluster would never recover and the play would always fail
+    - name: Enable shard allocation for the cluster
+      ansible.builtin.uri:
+        url: "{{ elasticsearch_http_protocol }}://localhost:{{ elasticstack_elasticsearch_http_port }}/_cluster/settings"
```
ivareri, line 76: Can we replace all occurrences of `localhost` with `elasticsearch_api_host`? 
ref https://github.com/NETWAYS/ansible-collection-elasticstack/issues/317 and https://github.com/NETWAYS/ansible-collection-elasticstack/pull/315

roles/elasticsearch/tasks/elasticsearch-rolling-upgrade.yml
```diff
@@ -0,0 +1,204 @@
...
+
+- name: Check for running Elasticsearch service
+  ansible.builtin.systemd:
+    name: elasticsearch
+  register: elasticsearch_running
+
+- name: Update stopped services right away
+  when:
+    - elasticsearch_running.status.ActiveState == "inactive"
+  block:
```
ivareri, line 26: I might be missing something, but why is this in a block while single instance upgrade is just two tasks?

roles/elasticsearch/tasks/elasticsearch-rolling-upgrade.yml
widhalmt: Yes, good catch! I'll do.

docs/role-elasticsearch.md
```diff
@@ -13,6 +13,7 @@ Role Variables
 --------------
 
 * *elasticsearch_node_types*: List of types of this very node. Please refer to [official docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-node.html) for details. (default: not set. allowed value: array of types)
++ *elasticsearch_nodename*': Node name of the Elasticsearch node. (default: Hostname of the node as seen by Ansible)
```
mkayontour, line 16: Default could be more specific, which variable is the "as seen by ansible" 
Could some of those:
- inventory_hostname 
- ansible_fqdn
- ansible_nodename
- ansible_hostname

roles/elasticsearch/tasks/elasticsearch-rolling-upgrade.yml
mkayontour, line 1: In general, I would add the os_family to os_specific tasks like the following:

```
    - name: Install Elasticsearch
      ansible.builtin.package:
        name: "{{ elasticsearch_package }}"
      when:
        - ansible_os_family == "Debian" or
          not elasticstack_full_stack | bool
```

Otherwise it could be very confusing if one install elasticsearch is skipped and the other not. 
Suggestion: 
`- name: [Debian] Install Elasticsearch`

mkayontour (commented):
I cannot really comment on the technical aspect of elasticsearch. 

But from the ansible perspective everything looks good to me. 

As I mentioned use some os_family information for os specific tasks, just to avoid confusion.

docs/role-elasticsearch.md
widhalmt: Thanks, I'll look it up in the code and will adjust it to avoid confusion

roles/elasticsearch/tasks/elasticsearch-rolling-upgrade.yml
widhalmt: Yes. I didn't like that as well. My idea was just that there's just a default variant and one for specific rpm setups. I'll make it more clear.

PR #255, opened 2023-09-05
Configure Galaxy in Molecule

### LINKED ISSUE #254: Parts of requirements files are ignored
By accident I found the following message in a jobs output:

```
  [WARNING]: The requirements file '/home/runner/work/ansible-collection-
  elasticstack/ansible-collection-
  elasticstack/molecule/elasticstack_default/requirements.yml' contains
  collections which will be ignored. To install these collections run 'ansible-
  galaxy collection install -r' or to install both at the same time run 'ansible-
  galaxy install -r' without a custom install path.
```

I'll look into it.

widhalmt:
This will force Molecule to run Ansible Galaxy twice. Once for roles, once for collection.

fixes #254

### DIFF
--- a/molecule/elasticsearch_cluster-oss/molecule.yml
+++ b/molecule/elasticsearch_cluster-oss/molecule.yml
@@ -1,6 +1,8 @@
 ---
 dependency:
   name: galaxy
+  options:
+    requirements-file: collections.yml
 driver:
   name: docker
 platforms:
--- a/molecule/elasticsearch_cluster-oss/molecule.yml
+++ b/molecule/elasticsearch_cluster-oss/molecule.yml
@@ -2,7 +2,7 @@
 dependency:
   name: galaxy
   options:
-    requirements-file: collections.yml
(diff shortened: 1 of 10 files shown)


widhalmt:
I'm not convinced that this change is really neccessary. It seems like the behaviour is exactly the same. My interpretation for now: First it installs the collections, then it runs again for roles. Since the both runs don't know of each other, the second issues the warning. So, since it's working as expected I'm ok with both - merging or closing.

Tests seem to fail because of timeouts at Rocky Linux repos.

afeefghannam89:
@widhalmt where is the collections.yml file, I do not see it? Did I misunderstand the goal of this PR? Can you please clarify more?

widhalmt:
Of course. I got the name wrong. 🤦🏽 I meant the `requirements.yml` in most Molecule scenarios.

afeefghannam89 (commented):
Thanks ;)

afeefghannam89 (approved):

PR #260, opened 2023-09-05
Fix/pipelines fix

### LINKED ISSUE #256: Make the container's name variable
We should make the container's name dynamic to prohibit collision on local GitHub action

### LINKED ISSUE #257: Remove failed_when by the elasticsearch service task
We should remove the failed_when on elasticsearch service start task. I as a user of this collection want to know that my elasticsearch service failed while starting it. I would not like to get failure by the API tasks after while and start with debugging to figure out at the end that the service did not start for x reason.
Forcing the service start task to not fail will not help or make a work around to the failed service.

afeefghannam89:
Fix #256 
Fix #257 
Fix #257


widhalmt (approved):
Looks good. This should help with parallel checks and speed up checks even more.

PR #269, opened 2023-09-07
Fix/debian pipeline

### LINKED ISSUE #258: Fix CI pipelines test on Debia10 and Rocky9
We have a collision problem on rocky9 while installing curl with the already available curl package. This issue is known and will be fixed from the maintainers. While we do not need to install curl on Rocky9, it is already available, we should remove this task.

On containerized Debian 10 systemd will not recognize elasticsearch searive without deamon-reload.

afeefghannam89:
Close #258 
This PR will fix the problem on debian the problem on rocky is already fixed in another branch, but GitHub did not close the issue that time, I do not know why.

### DIFF
--- a/.github/workflows/test_roles_pr.yml
+++ b/.github/workflows/test_roles_pr.yml
@@ -51,13 +51,12 @@ jobs:
       max-parallel: 2
       matrix:
         distro:
-          #- rockylinux8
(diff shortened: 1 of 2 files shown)


widhalmt (approved):
Good catch! LGTM

PR #270, opened 2023-09-07
Add "extra config" to `kibana.yml`

### LINKED ISSUE #208: enhancement: additional configuration for kibana.yml
For my perspective it is a good enhancement to add the possibility to add additional parameters to the kibana.yml.

e.g. in our case the following parameters needed.
telemetry.enabled: false
monitoring.ui.ccs.enabled: false

I think a good approach are a generic way. 
For example have a variable like this and loop over it in the jinja template file

```
kibana.additional.parameter:
  telemetry.enabled: false
  monitoring.ui.ccs.enabled: false
  other_parameters: value
```

widhalmt:
fixes #208


dnssmr (approved):
Looks good to me!

PR #271, opened 2023-09-08
* Remove cache after all roles run

### LINKED ISSUE #272: Elasticsearch cluster state is yellow
Elasticsearch cluster state is yellow on centos8, because of not enough space to allocate shards. There was a fix for rocky 8 and Debian, but here Elasticsearch has problem with space after the installation of Kibana, Logstash and Beats, not only after installing Elasticsearch

afeefghannam89:
We should free up space after the roles run to have green cluster state in impotence step and enable Elasticsearch to allocate all indices.

Close #272

### DIFF
--- a/.github/workflows/test_roles_pr.yml
+++ b/.github/workflows/test_roles_pr.yml
@@ -51,9 +51,8 @@ jobs:
       max-parallel: 2
       matrix:
         distro:
-          - centos8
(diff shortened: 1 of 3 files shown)


widhalmt (approved):
Everything that speeds up tests, helps. ;-) Thanks.

PR #273, opened 2023-09-08
Feature/usermanagement

pdolinic:
Hi, this PR creates a new role that allows dynamically updating roles and users. From here nearly any API request can be automated in a similar way.

- Please feel free to transform this as a **draft** or other **to reuse any code parts,** snippets or ideas as you wish, and add them at any places you want, without the need to give credit. I am happy to serve the Open Source community.

- The test are not in molecule format, but included, and usage is hopefully explanatory enough out of the documentation.
- Currently there are 2 test vaults with inputs for 1) elastic-api password and 2) elastic users to be added. The test secrets should be visible that they are encrypted for both cases. It is possible like this to extend this to anything else in this style, for example the CA password, or roles nobody should need to know something about.

I hope it is of use to someone else


martialblog:
Closing this PR. Keeping the branch... for now.

PR #274, opened 2023-09-12
Use common variable for TLS

### LINKED ISSUE #248: Metricbeat can't automatically connect
When running the collection with defaults (only activating Metricbeat) it fails, because the variable `elasticsearch_http_security` isn't set for Metricbeat hosts. This looks like it's some legacy code from before we built the collection. It's possible that Auditbeat suffers the same problem.

I'll investigate and I hope I can provide a solution.

widhalmt:
fixes #248

The `metricbeat.yml` and `auditbeat.yml` templates have code for both connecting to Logstash or Elasticseach. They used different variables to determine whether to turn on TLS or not. That resulted in a missing default and therefore Beats setup breaking when connecting to Elasticseach. I changed the variable that's queried to determine whether to turn on TLS or not.

This is something between a workaround and a partial solution. While it makes more sense to use a role related variable for both Logstash and Elasticsearch output (other than before) it's still not the goal we want.

So overhauling the whole process of determining whether we need TLS or not is due. I'll start a discussion and we need to adjust all roles to the solution we find. Now there are similarities but it's not exactly the same in every role.


dnssmr:
Is the documentation still valid with these changes?

widhalmt:
... I hope, I didn't miss anything. But I'll have a second look.

DanOPT:
@widhalmt Can I push documentation for `elasticsearch_http_security` and `elasticsearch_http_protocol` to this PR. I can check and add documentation for `beats_security`, if required, too.

widhalmt:
@DanOPT of course! Go ahead.

widhalmt:
@lcndsmr sorry for the eternal wait... yes, the documentation already has the new variables. So this fix actually makes code matching the docs better.

widhalmt:
@lcndsmr could you recheck?

PR #277, opened 2023-09-28
Set Debian 11 as new default distro for molecule

### LINKED ISSUE #245: Change default distro for Molecule from CentOS 7 to something more current
While using CentOS 7 for Molecule was a sensitive choice in "ye goode olde days"(tm) we should move to something else. I suggest, we decide on which distro to use considering @geerlingguy ' s roadmap for containers.

widhalmt:
We had CentOS 7 long enough so I thought it might be nice to go with Debian instead. This PR will not change much in the behaviour because usually we overwrite the default. So this is just in case and to stay up to date.

fixes #245

### DIFF
--- a/molecule/beats_default/molecule.yml
+++ b/molecule/beats_default/molecule.yml
@@ -4,8 +4,8 @@ dependency:
 driver:
   name: docker
 platforms:
-  - name: beats_default_${MOLECULE_DISTRO:-centos7}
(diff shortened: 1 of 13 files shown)


martialblog (approved):
Change makes sense. Bookworm would be nice, but the conservative choice is Bullseye

widhalmt:
Checks are arbitrarily failing, again. I try to tackle it in #278

widhalmt:
We need to merge #279 for checks to succeed

PR #279, opened 2023-09-28
Remove redundant restart

### LINKED ISSUE #141: Restarting Elasticsearch after the first start is not ideal
We flush handlers in security file yaml after the first start of the new initialized cluster. That lead to directly restart of Elasticsearch after the first start. This is maybe not ideal. We can put a when condition to adjust this behavior. 

Also, we put two tasks to start Elasticsearch. The first one in security file yaml and the next one in main.yml. This led to two changes in Ansible output. What about putting when condition to not trigger the second one if security is enabled?

What do you think @widhalmt?

### LINKED ISSUE #194: Fix full stack workflow on all containers
Full stack work flow does not work as expected on some containers. We should fix this or remove these containers from testing.

### LINKED ISSUE #278: Redundant restarts
We have handlers restarting services and thus creating lag that can lead to timeouts in our checks.

I guess, we can reduce the count of restarts, especially for Elasticsearch and speed up the checks dramatically.

One issue I found so far is that in #137 we agreed on not restarting the cluster after the change, but we still notify the handler.

I'll look into the roles and try to remove any redundant restart.

widhalmt:
Restarting Elasticsearch takes quite a while and may lead to connection issues as well as sync issues. So keeping restarts to a minimum is important. These changes will make sure that, even when the `Restart Elasticsearch` handler is notified, it will only restart if Elasticsearch was running before. If there's a fresh start (after reconfiguration) we don't need to restart again.

Same goes for Logstash and Kibana. Some restarts of these tools happen fairly fast. But others (like after fresh installs or updates) will trigger internal jobs that should not be intercepted by another restart.

Beats restart very fast and as far as I know there's not a big downside to restarting them right after the first start so I didn't include them in the change.

Additionally, this PR will make sure some tasks in `verify.yml` of the full stack are only run when the service to be checked is actually running on this node. This helps with spreading services over nodes to save ressources.

Since GitHub hosted runners are quite low on ressources we can't run every service on every node in a cluster setup anymore. So this PR will make sure that only Elasticsearch runs everywhere and the others are spread out.

Caches get cleared after every role in during a Molecule test. This helps with saving ressources, too.

Elasticsearch still won't sync all shards due to full volumes, the watermarks for Elasticseach are set to extremely high volumes so that the cluster can at least get into sync.

fixes #278
fixes #141 
fixes #194

### DIFF
--- a/roles/beats/tasks/main.yml
+++ b/roles/beats/tasks/main.yml
@@ -82,3 +82,11 @@
 - name: Import Metricbeat tasks
   ansible.builtin.import_tasks: metricbeat.yml
   when: beats_metricbeat | bool
+
+# Free up some space to let elsticsearch allocate replica in GitHub Action
+- name: Remove cache # noqa: risky-shell-pipe
+  ansible.builtin.shell: >
+    if test -n "$(ps -p $$ | grep bash)"; then set -o pipefail; fi;
+    rm -rf /var/cache/*
+  changed_when: false
+  when: ansible_virtualization_type == "container" or ansible_virtualization_type == "docker"
--- a/roles/beats/tasks/main.yml
+++ b/roles/beats/tasks/main.yml
@@ -84,9 +84,8 @@
   when: beats_metricbeat | bool
 
 # Free up some space to let elsticsearch allocate replica in GitHub Action
-- name: Remove cache # noqa: risky-shell-pipe
-  ansible.builtin.shell: >
-    if test -n "$(ps -p $$ | grep bash)"; then set -o pipefail; fi;
+- name: Remove cache
+  ansible.builtin.command: >
     rm -rf /var/cache/*
   changed_when: false
   when: ansible_virtualization_type == "container" or ansible_virtualization_type == "docker"
--- a/roles/elasticsearch/tasks/elasticsearch-security.yml
+++ b/roles/elasticsearch/tasks/elasticsearch-security.yml
@@ -451,6 +451,44 @@
   retries: 20
   delay: 10
 
+- name: Work around low ressources on CI/CD nodes
+  when: ansible_virtualization_type == "container" or ansible_virtualization_type == "docker"
+  block:
+    # Free up some space to let elsticsearch allocate replica in GitHub Action
+    - name: Remove cache # noqa: risky-shell-pipe
+      ansible.builtin.shell: >
+        if test -n "$(ps -p $$ | grep bash)"; then set -o pipefail; fi;
+        rm -rf /var/cache/*
+      changed_when: false
+
+    - name: Set persistent watermarks to very high values in Docker # noqa: risky-shell-pipe
+      ansible.builtin.shell: >
+        if test -n "$(ps -p $$ | grep bash)"; then set -o pipefail; fi;
+        curl
+        -k
+        -X PUT
+        "{{ elasticsearch_http_protocol }}://elastic:{{ elasticstack_password.stdout }}@localhost:9200/_cluster/settings"
+        -H 'Content-Type: application/json' -d
+        '
+        {
+          "persistent": {
+            "cluster.routing.allocation.disk.watermark.low": "97%",
+            "cluster.routing.allocation.disk.watermark.high": "98%",
+            "cluster.routing.allocation.disk.watermark.flood_stage": "99%",
+            "cluster.routing.allocation.disk.watermark.flood_stage.frozen": "99%"
+          }
+        }
+        '
+      changed_when: false
+      register: debugme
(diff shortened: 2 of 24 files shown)


widhalmt:
Seems like the Elasticsearch cluster still has problems syncing. I'll withdraw the request for review and get back to Draft.

roles/beats/tasks/main.yml
```diff
@@ -82,3 +82,11 @@
...
   ansible.builtin.import_tasks: metricbeat.yml
   when: beats_metricbeat | bool
+
+# Free up some space to let elsticsearch allocate replica in GitHub Action
+- name: Remove cache # noqa: risky-shell-pipe
+  ansible.builtin.shell: >
+    if test -n "$(ps -p $$ | grep bash)"; then set -o pipefail; fi;
+    rm -rf /var/cache/*
+  changed_when: false
+  when: ansible_virtualization_type == "container" or ansible_virtualization_type == "docker"
```
widhalmt, line 92: This shouldn't need `shell` but could do with `cmd` as well. Could save some time during testing. But I was lazy and just copied it. Would it make sense to replace it? (Note that there are several occurences of this task in the whole collection)

roles/elasticsearch/tasks/elasticsearch-security.yml
```diff
@@ -410,6 +435,58 @@
...
+            "cluster.routing.allocation.disk.watermark.high": "98%",
+            "cluster.routing.allocation.disk.watermark.flood_stage": "99%",
+            "cluster.routing.allocation.disk.watermark.flood_stage.frozen": "99%"
+          }
+        }
+        '
+      changed_when: false
+      no_log: "{{ elasticstack_no_log }}"
+      when:
+        - elasticstack_password.stdout is defined
```
widhalmt, line 485: I'm totally torn between keeping this and reworking it yet again. If I rework it, there should be a clean way of setting watermarks by variables based on a template and the `uri` module.

Here's what I thought so far:

Contra reworking:
* Setting watermarks isn't very common (which is a euphemism) in systems aside from this test case with extremely low ressources
* This is a workaround for something (GitHub runners with low ressources) we might find a better solution soon and just replace it all together
* Since checking current settings before setting them would introduce another level of complexity it would most likely result in hammering in the same settings with every run
* Unsetting to default would be another thing that might not be worth the extra effort
* Setting Elasticsearch settings would be something that might be really helpful in other circumstances. So even when at first glance, a module wouldn't do much more than replace a few options of `uri` I guess it would be really beneficial to have one that could check settings before setting them again. And be able to undo it. But this would clearly be out of the scope of this PR. (Had this in the "PRO" section, but actually I think this means "keep the humble workaround and open an issue to create a module")

Pro reworking:
* Everything you can customize but still has sensible defaults (e.g. those from the vendor) is a bonus for edge cases
* Using `curl` in `shell` is something @dnssmr put a lot of effort in to get rid off. And I don't want to re-introduce it out of laziness

dnssmr: How much work do you think reworking this would be? And how probable is it, that we introduce another solution soon?

roles/beats/tasks/main.yml
dnssmr: If this command works in `ansible.builtin.command` and is used more than once, please use it. `shell` is considerably slower.

widhalmt: Thanks for your input. I'll test and rework it.

widhalmt: I changed it. Could you have another look, @lcndsmr

roles/elasticsearch/tasks/elasticsearch-security.yml
widhalmt: Writing custom modules is still on my to-do list. So it's hard for me to tell how much work a separate module would be. But I could imagine, having a generic module to get, set and update arbitraty settings in Elasticsearch could be quite beneficial. I'd like to see it soon but we already have some modules to write on our table.

Rewriting it, to simply use `uri` with more or less hardcoded settings wouldn't be that much of a hassle. It's just the question if it's beneficial enough to do just for the sake of getting rid of `curl` in a makeshift solution.

roles/beats/tasks/main.yml
lcndsmr: Looks good, thank you!

roles/elasticsearch/tasks/elasticsearch-security.yml
lcndsmr: Then I#d leave it as is for now!

PR #284, opened 2023-10-11
Remove all max-parallel from GitHub workflows

### LINKED ISSUE #283: Deactivate throtteling to speed up testing
Like @afeefghannam89 told me once, GitHub workflows use distinct VMs and should not overlap each other. So it should be save to remove our throtteling without affecting tests. This could greatly speed up checks.

widhalmt:
This will need #279 to be merged before tests can pass.

fixes #283


lcndsmr (approved):
looks good

PR #285, opened 2023-10-11
Replace quotations for passwords with special characters

### LINKED ISSUE #282: elasticsearch_tls_key_passphrase containing $ not working
`elasticsearch_tls_key_passphrase` with a `$` character is not working. This is most likely due to using echo with `"` quotation. This will lead to parameter expansion, while using `'` wont. 

```
~ >>> echo "test$var@test"                                                                                                                                                                
test@test
```
vs
```
~ >>> echo 'test$var@test'                                                                                                                                                                
test$var@test
```

This is done several places in [elasticsearch-keystore.yml](https://github.com/NETWAYS/ansible-collection-elasticstack/blob/main/roles/elasticsearch/tasks/elasticsearch-keystore.yml), like [here](https://github.com/NETWAYS/ansible-collection-elasticstack/blob/62bf591a20234394a76ad43232bd846cb3206841/roles/elasticsearch/tasks/elasticsearch-keystore.yml#L42C48-L42C48)

widhalmt:
Using special characters like `$` in some passwords breaks the role. This PR introduces a test (by setting passwords with special characters) and introduces a fix to handle them correctly.

fixes #282

### DIFF
--- a/molecule/elasticstack_default/converge.yml
+++ b/molecule/elasticstack_default/converge.yml
@@ -12,6 +12,7 @@
   vars:
     elasticsearch_jna_workaround: true
     elasticsearch_disable_systemcallfilterchecks: true
+    elasticsearch_tls_key_passphrase: 'test$var@test' # to test escaping
     elasticstack_release: "{{ lookup('env', 'ELASTIC_RELEASE') | int}}"
     elasticsearch_heap: "1"
     elasticstack_full_stack: true
--- a/roles/elasticsearch/tasks/elasticsearch-keystore.yml
+++ b/roles/elasticsearch/tasks/elasticsearch-keystore.yml
@@ -13,7 +13,7 @@
 - name: Set bootstrap password # noqa: risky-shell-pipe
   ansible.builtin.shell: >
     if test -n "$(ps -p $$ | grep bash)"; then set -o pipefail; fi;
-    echo "{{ elasticsearch_bootstrap_pw }}" |
+    echo '{{ elasticsearch_bootstrap_pw }}' |
     /usr/share/elasticsearch/bin/elasticsearch-keystore
     add -x 'bootstrap.password'
   when: "'bootstrap.password' not in elasticsearch_keystore.stdout_lines"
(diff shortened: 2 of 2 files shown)


ivareri:
I haven't tested, but is there a reason not to use  [`ansible.builtin.command`](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/command_module.html) instead of `shell` for this?

Something like this might work
```YAML
- name: Set bootstrap password
  ansible.builtin.command: /usr/share/elasticsearch/bin/elasticsearch-keystore add -x 'bootstrap.password'
  args:
    stdin: {{ elasticsearch_bootstrap_pw }}
  when: "'bootstrap.password' not in elasticsearch_keystore.stdout_lines"
```

widhalmt:
> I haven't tested, but is there a reason not to use [`ansible.builtin.command`](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/command_module.html) instead of `shell` for this?
> 
> Something like this might work
> 
> ```yaml
> - name: Set bootstrap password
>   ansible.builtin.command: /usr/share/elasticsearch/bin/elasticsearch-keystore add -x 'bootstrap.password'
>   args:
>     stdin: {{ elasticsearch_bootstrap_pw }}
>   when: "'bootstrap.password' not in elasticsearch_keystore.stdout_lines"
> ```

To be honest, I don't see a reason. Maybe the contributer back then had an idea I do not recognize. I guess, in the current implementation, command should be fine and make things even faster.

In the long term, it might be a good idea to replace these tasks with a custom module, though.

widhalmt:
@ivareri I guess it was that the original contributor (like me) didn't think of the possibility to implement it with `stdin`.

I still think this would fit into another PR so not to change too many things at once. Maybe a module would be even better.

tbauriedel (approved):

afeefghannam89 (changes requested):
The suggestion of @ivareri with command module is the best one because:
- There is no Shell in between when writing the password in the store. when auditd + execve enabled, the password will be plain in the logs.
- When the password has single quote will not be a problem. stdin accept everything

afeefghannam89:
Complated with #470

PR #288, opened 2023-10-15
Elasticsearch config

ivareri:
Minor tweaks and fixes to Elasticsearch configuration file

### DIFF
--- a/docs/role-elasticsearch.md
+++ b/docs/role-elasticsearch.md
@@ -39,6 +39,7 @@ This variable activates a workaround to start on systems that have certain harde
 * *elasticsearch_http_port*: The port to bind for HTTP client communication
 * *elasticsearch_transport_port*: The port to bind for communication between nodes
 * *elasticsearch_seed_hosts*: Set elasticsearch seed hosts
+* *elasticsearch_security_enrollment*: Controls enrollment (of nodes and Kibana) to a local node that’s been autoconfigured for security.
 
 These variables are identical over all our elastic related roles, hence the different naming schemes.
 
--- a/docs/role-elasticsearch.md
+++ b/docs/role-elasticsearch.md
@@ -38,6 +38,7 @@ This variable activates a workaround to start on systems that have certain harde
 * *elasticsearch_ssl_verification_mode*: Defines how to verify the certificates presented by another party in the TLS connection
 * *elasticsearch_http_port*: The port to bind for HTTP client communication
 * *elasticsearch_transport_port*: The port to bind for communication between nodes
+* *elasticsearch_seed_hosts*: Set elasticsearch seed hosts
 
 These variables are identical over all our elastic related roles, hence the different naming schemes.
 
--- a/docs/role-elasticsearch.md
+++ b/docs/role-elasticsearch.md
@@ -36,6 +36,8 @@ This variable activates a workaround to start on systems that have certain harde
 
 * *elasticsearch_jna_workaround*: Activate JNA workaround. (default: `false`)
 * *elasticsearch_ssl_verification_mode*: Defines how to verify the certificates presented by another party in the TLS connection
+* *elasticsearch_http_port*: The port to bind for HTTP client communication
+* *elasticsearch_transport_port*: The port to bind for communication between nodes
 
 These variables are identical over all our elastic related roles, hence the different naming schemes.
 
--- a/docs/role-elasticsearch.md
+++ b/docs/role-elasticsearch.md
@@ -35,6 +35,7 @@ Role Variables
 This variable activates a workaround to start on systems that have certain hardening measures active. See [Stackoverflow](https://stackoverflow.com/questions/47824643/unable-to-load-jna-native-support-library-elasticsearch-6-x/50371992#50371992) for details and logmessages to look for. **WARNING**: This will change your `/etc/sysconfig/elasticseach`or `/etc/default/elasticsearch` file and overwrite `ES_JAVA_OPTS`. See this [issue](https://github.com/netways/ansible-role-elasticsearch/issues/79) for details.
 
 * *elasticsearch_jna_workaround*: Activate JNA workaround. (default: `false`)
+* *elasticsearch_ssl_verification_mode*: Defines how to verify the certificates presented by another party in the TLS connection
 
 These variables are identical over all our elastic related roles, hence the different naming schemes.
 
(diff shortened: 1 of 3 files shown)


docs/role-elasticsearch.md
```diff
@@ -35,6 +35,11 @@ Role Variables
 This variable activates a workaround to start on systems that have certain hardening measures active. See [Stackoverflow](https://stackoverflow.com/questions/47824643/unable-to-load-jna-native-support-library-elasticsearch-6-x/50371992#50371992) for details and logmessages to look for. **WARNING**: This will change your `/etc/sysconfig/elasticseach`or `/etc/default/elasticsearch` file and overwrite `ES_JAVA_OPTS`. See this [issue](https://github.com/netways/ansible-role-elasticsearch/issues/79) for details.
 
 * *elasticsearch_jna_workaround*: Activate JNA workaround. (default: `false`)
+* *elasticsearch_ssl_verification_mode*: Defines how to verify the certificates presented by another party in the TLS connection
+* *elasticsearch_http_port*: The port to bind for HTTP client communication
```
widhalmt, line 39: Please be aware that we're using `elasticstack_elasticsearch_http_port` inside the roles to connect to Elasticsearch.

I started a conversation whether we should have these "global" variables in an extra role or not. I wanted to use a "global" so we have the same variable for every service that connects to Elasticsearch. Rewriting it into a local version is ok but then we need to set it to the global variant. So either please go with the global one or add some code to tackle that.

docs/role-elasticsearch.md
```diff
@@ -35,6 +35,11 @@ Role Variables
...
+* *elasticsearch_transport_port*: The port to bind for communication between nodes
```
widhalmt, line 40: No objection to that. Transport is completely local to Elasticsearch.

docs/role-elasticsearch.md
```diff
@@ -35,6 +35,11 @@ Role Variables
...
+* *elasticsearch_seed_hosts*: Set elasticsearch seed hosts
```
widhalmt, line 41: I know, we have some variables of type string to fill arrays. Personally I'd prefer to have a list in Ansible that will fill the array but I totally get that this needs extra code and still you can't control which interface is used etc. So I'm ok with it.

roles/elasticsearch/defaults/main.yml
```diff
@@ -47,6 +47,7 @@ elasticstack_ca_expiration_buffer: 30
 elasticsearch_cert_expiration_buffer: 30
 elasticstack_ca_will_expire_soon: false
 elasticsearch_cert_will_expire_soon: false
+elasticsearch_ssl_verification_mode: none
```
widhalmt, line 50: Please set it to `full`. I know, TLS can be a pest to implement, but we want to have every user create an exception on purpose and keep the defaults as secure as possible. I just saw we had `none` hardcoded. So, let's say, you found a bug and fixed it. :-)

roles/elasticsearch/templates/elasticsearch.yml.j2
```diff
@@ -48,14 +61,18 @@ bootstrap.system_call_filter: false
 {% endif %}
 {% if elasticstack_variant == "elastic" %}
 xpack.ml.enabled: {{ elasticsearch_ml_enabled }}
+{% if elasticstack_release == 7 %}
 xpack.monitoring.collection.enabled: {{ elasticsearch_monitoring_enabled }}
+{% endif %}
```
widhalmt, line 66: It is deprecated since 7 but it still works in 8, doesn't it? I haven't seen any error message but maybe I've just missed it.

roles/elasticsearch/templates/elasticsearch.yml.j2
```diff
@@ -48,14 +61,18 @@ bootstrap.system_call_filter: false
...
 {% if elasticsearch_security | bool %}
 xpack.security.enabled: true
+{% if elasticsearch_security_enrollment is defined %}
+xpack.security.enrollment.enabled: {{ elasticsearch_security_enrollment }}
+{% endif %}
```
widhalmt, line 71: Honestly, I haven't dealt with that. I need to dig into it a bit more before I can approve. Sorry, that's totally my fault.

widhalmt (changes requested):
Thanks again for the contribution. I have a few changes to request.

Some comments are just that - comments. Please just close the conversation when you read them and agree.

And I have to dig into that auto-enrollment part on my side. Maybe I find a colleague that has more experience with it.

roles/elasticsearch/defaults/main.yml
ivareri: Fixed. 

I set it to none in order to keep the current behavior, but I agree full is a better default.

roles/elasticsearch/templates/elasticsearch.yml.j2
ivareri: As I understand it the setting is simply ignored in 8. 

While it doesn't interfere with normal operations, if something unrelated breaks I fear it might send someone troubleshooting the installation down the wrong path. The install script leaves a small configuration file, with no comments. That would lead me to think every setting is there for a reason, and I'd rather not leave any (unintentional) red herrings

docs/role-elasticsearch.md
ivareri: Ohh, I missed that one. Changed so it uses `elasticstack_elasticsearch_http_port`.

roles/elasticsearch/templates/elasticsearch.yml.j2
ivareri: I think this is a setting that you usually don't change, and it's fine to leave it unset. However, I had one installation where it was needed. I figured not setting a default for it and having the setting in an if clause would allow it's use in the few cases where needed, while not causing any harm when not needed.

lcndsmr: I think its totally fine to add this!

roles/elasticsearch/templates/elasticsearch.yml.j2
widhalmt: You're right. The change is ok with me. Besides, putting more comments into the generated files is definitely a goal for later development, good point!

widhalmt (approved):
Just revisited the PR. Thank you and sorry for the waiting time. Approved.

PR #292, opened 2023-11-10
Set passphrase for beats tls key

### LINKED ISSUE #291: Default beats input doesn't use passphrase for TLS key
This is a potential problem in two ways:

* It's a security issue to have an unprotected key
* The configuration has no option to set a password. So even when you create your own key, you can't use it with the collection

widhalmt:
Side effect: Includes extra changes for listing names of pipelines in task names

fixes #291


widhalmt:
Seems like setting the passphrase with default options works with Logstash 8 but not Logstash 7.

PR #297, opened 2024-01-15
Fix extra fields for Beats

### LINKED ISSUE #296: [Bug]: Extra fields in Filebeat inputs not working

_No response_

### Describe the bug

When using die `fields` setting in `filebeat_log_inputs` you don't get the whole line in `fields` within `filebeat.yml`. Only the field name but not the value.

Thanks, @tbumke for reporting this bug.

### How to recreate the bug?

Use variables like this:

```
filebeat_log_inputs:
  messages:
    name: messages
    paths:
      - /var/log/messages
    fields:
      index_suffix: system
```
The value `system` will not be written to `filebeat.yml`.

widhalmt:
fixes #296

### DIFF
--- a/molecule/beats_peculiar/converge.yml
+++ b/molecule/beats_peculiar/converge.yml
@@ -18,6 +18,8 @@
           pattern: '^[[:space:]]+(at|\.{3})[[:space:]]+\b|^Caused by:'
           negate: false
           match: after
+        fields:
+          testfield: thisisatest
     elasticsearch_jna_workaround: true
     elasticsearch_disable_systemcallfilterchecks: true
     elasticstack_full_stack: false
--- a/molecule/beats_peculiar/verify.yml
+++ b/molecule/beats_peculiar/verify.yml
@@ -26,3 +26,7 @@
       when:
         - beats_filebeat_version.stdout.find('8.4.1') == -1
         - elasticstack_release == 8
+
+    - name: Check for fields being set
+      ansible.builtin.command: >
+        pcregrep -M 'fields:(\n).*testfield: thisisatest' /etc/filebeeat/filebeat.yml
--- a/roles/beats/templates/filebeat.yml.j2
+++ b/roles/beats/templates/filebeat.yml.j2
@@ -12,7 +12,7 @@ filebeat.inputs:
   fields:
 {% if beats_fields is defined %}
 {% for field in beats_fields %}
-    {{ field }}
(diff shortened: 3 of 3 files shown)


widhalmt:
I still have comments with a more sophisticated check. To be honest, today feels like I'm cursed. Couldn't type a single straight line. Since this fix is somewhat time critical I go with a manual check this time. If you're reviewing you're very welcome to think this approach is wrong and come up with a better solution. I for my part tried my best for today. Maybe I'll do something better later but by then the fix should be integrated for a long time already.

widhalmt:
Helpful for review:
* Added two fields to `converge.yml` of the `beats_peculiar` molecule scenario
* In the output of the `verify` stage of the `beats_peculiar` scenario you should see the fields added to `fielbeat.yml`
* Filebeat needs to run because there's a chance that it refuses to start or crashes when the configuration contains errors

widhalmt:
Needs #300 to be fixed before all neccessary checks can run.

tbauriedel:
@widhalmt 
Since #300 is implemented now, you will also have to merge these changes into your PR that the pipelines can start.

widhalmt:
@tbauriedel thanks. Waiting for the merge queue to complete.

tbauriedel:
@widhalmt I have updated the verification for the fields for beats_peculiar.

Correct me if I am wrong. But currently no fields are defined inside the elasticstack_default converge?
So no test necessary?

widhalmt:
Exactly. I'm trying to keep `elasticstack_default` to defaults as much as possible. So we make sure you can use the collection with defaults without digging through every option.

molecule/elasticstack_default/verify.yml
```diff
@@ -135,3 +135,11 @@
...
       with_inventory_hostnames: all
       when: groups['elasticsearch'] | length > 1
+
+    - name: Fetch filebeat.yml
+      ansible.builtin.command: cat /etc/filebeat/filebeat.yml
+      register: filebeatyml
+
+    - name: Show filebeat.yml
+      ansible.builtin.debug:
+        var: filebeatyml.stdout_lines
```
tbauriedel, line 145: As you want to keep this workflow as default as possible and no fields are defined, the task here is not needed I think.
Regardless of this, you will probably look at the debug the first two times and then forget about it.

```suggestion
```

tbauriedel (changes requested):

tbauriedel (approved):

PR #306, opened 2024-01-25
Fix apt key handling

zerwes:
avoid warning on debian 12
```
W:https://artifacts.elastic.co/packages/8.x/apt/dists/stable/InRelease: Key is stored in legacy trusted.gpg keyring (/etc/apt/trusted.gpg), see the DEPRECATION section in apt-key(8) for details
```

### DIFF
--- a/roles/repos/defaults/main.yml
+++ b/roles/repos/defaults/main.yml
@@ -1,5 +1,6 @@
 ---
 # defaults file for elastic-repos
+elasticstack_repo_key: https://artifacts.elastic.co/GPG-KEY-elasticsearch
 elasticstack_release: 8
 elasticstack_full_stack: true
 elasticstack_variant: elastic
--- a/roles/repos/tasks/debian.yml
+++ b/roles/repos/tasks/debian.yml
@@ -6,10 +6,16 @@
       - gpg-agent
     state: present
 
-- name: Ensure Elastic Stack key is available (Debian)
+- name: Ensure Elastic Stack key is configured (Debian legacy)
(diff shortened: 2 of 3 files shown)


widhalmt:
Hi,

Thanks so much for your contribution!

The checks fail due to idempotency tests. Could it be that there's a newer `apt_key` module that will put it in the right place on Ubuntu? It looks like one task is removing it and the other is putting it back again.

zerwes:
Hello @widhalmt 
Oh, sorry, I have to admit: my mistake.
Usually, while managing all hosts mainly via ansible, the apt warning is hidden. I just stumbled over it while testing some things manually. And I just applied an older fix I had developed for other repositories and took the short way, **yolo**, and just opened the PR.
But things develop and meanwhile it seems the `apt_key` task is reading keys from `/etc/apt/trusted.gpg.d/` too and removing them if requested ... so for surely the idempotency check will fail, as the key is removed and re-added.
Well meanwhile, the `apt_key` command ist deprecated in ansible too, so, while thinking about a clean solution, I stumbled over the next issue: storing keys in `/etc/apt/trusted.gpg.d/` removes the apt warning, but the keys are still trusted in a global scope. The new an best way to do it is to restict the key trust just for the repo in question. The `deb822_repository` is available for `ansible >= 2.15`, so not yet ready for collections in a wider use.
SO I will close this PR and reopen another one (considering the idempotency check)
Sorry for the trouble, thank you for your hint and until the next PR targeting this issue.
Greetings
Klaus

Issue #313, opened 2024-02-07
[Bug]: Rework version selection, especially in beats


widhalmt:
### Please try to fill out as much of the information below as you can. Thank you!

- [X] Yes, I've searched similar issues on GitHub and didn't find any.

### Describe the bug

If you check the code parts that install packages you will find that Beats have twice as many tasks as others. That's because they have support for `latest` in `version`.

So either we build something similar in all the roles (don't like the idea) or we get rid of the option to set `version` to `latest`. When building a solution, please consider #250 where we add code to first determine which version to install and then use this for all other packages.

### How to recreate the bug?

Nothing really to recreate. Just review the code.

PR #314, opened 2024-02-07
Enforce repos only with full stack

### LINKED ISSUE #312: [Bug]: Forcfully enabling repository breaks install task on some systems

_No response_

### Describe the bug

We have a mandatory `enable_repo` in the Logstash RedHat installation task file. This is fine when you use the included repo role. But if you have e.g. RedHat Satellite to manage your repositories you might not have a repository with this exact name. So the install task breaks.

So either we make this task into two different ones or we remove the "`enable_repo` part and find another way to satisfy users who want to disable the repository in case they would accidentally update.

### How to recreate the bug?

* Rename Elastic Repository
* Disable `repos` role
* Run Logstash role with installation

widhalmt:
RPM installation automatically enabled the Elastic Stack repositories this collection will add to a host. But if you have another tool to manage your repositories, they might have other ids and names.

So I added separate installation tasks. If `elasticstack_full_stack` is set to `true`, everything stays as it was. The task will forcefully enable the repository and then install the package.

But if `elasticstack_full_stack` is set to `false` it will just ignore the repository and expect the package to be available.

While working on this I found a different approach to choosing the correct version in the `beats` role. I couldn't afford to fix it all in a single PR so I opened #313 for this.

fixes #312

### DIFF
--- a/roles/beats/tasks/auditbeat.yml
+++ b/roles/beats/tasks/auditbeat.yml
@@ -10,13 +10,22 @@
       string if elasticstack_version is defined else '') |
       replace(' ', '')
       }}
-- name: Install Auditbeat - rpm
+
+- name: Install Auditbeat - rpm - full stack
   ansible.builtin.package:
     name: "{{ beats_auditbeat_package }}"
     enablerepo:
       - 'elastic-{{ elasticstack_release }}.x'
   when:
     - ansible_os_family == "RedHat"
+    - elasticstack_full_stack | bool
+
+- name: Install Auditbeat - rpm - standalone
+  ansible.builtin.package:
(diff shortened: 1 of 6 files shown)


tbauriedel:
You could combine the variables to have one. But thats also just a "hacky workaround".
`elastic_repo: "{{ filename }}-{{ version }}"`

I don't like the whole thing right now either. It's all very hacky. But I can't think of a better solution at the moment.
Solving the whole thing using a variable would be the cleanest way.

widhalmt:
That can only work with another extra task assembling the name when the value's empty. Because, if we allow to have arbitraty repo names then we can't make any assumptions about the nomenclature.

tbauriedel (approved):

PR #315, opened 2024-02-09
Use elasticsearch_network_host for connection tests if defined

ivareri:
When `elasticsearch_network_host` is set to something not including localhost, all connection tests fails. 

This uses `elasticsearch_network_host` for connections if it is defined, and falls back to localhost if not. 

(There might be more places this change should be made, but the role works for me with this fix)

### DIFF
--- a/roles/elasticsearch/tasks/elasticsearch-security.yml
+++ b/roles/elasticsearch/tasks/elasticsearch-security.yml
@@ -391,7 +391,7 @@
 
 - name: Check for API with bootstrap password
   ansible.builtin.uri:
-    url: "{{ elasticsearch_http_protocol }}://localhost:{{ elasticstack_elasticsearch_http_port }}"
+    url: "{{ elasticsearch_http_protocol }}://{{ elasticsearch_network_host|default('localhost') }}:{{ elasticstack_elasticsearch_http_port }}"
     user: elastic
     password: "{{ elasticsearch_bootstrap_pw }}"
     validate_certs: false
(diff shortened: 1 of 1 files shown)


widhalmt:
Thanks for your contribution! The more we can configure, the better. :-)

widhalmt (changes requested):
Clean and neat. Thank you.

`elasticsearch_network_host` is a string consisting of different entries in a single string. So this will fail if you set more than one entry for it. I'm afraid we need to rework the logic of this variable so we can still have several values and have your change work.

widhalmt:
If you want, I can propose a change so the variable will work as a list instead of a string. Then you can fix your PR and we can merge it.

widhalmt:
When looking at the code, this seems to be even more complicated than I thought. What if someone uses macros like `_local_` or `_site_` in their `elasticsearch_network_host` configuration? `curl` and `uri` can't use that. Any idea?

ivareri:
Well duh, I had a feeling this was too neat of a fix. 

I agree, `elasticsearch_api_host` is ugly, but probably the best solution. I'll update the PR.

ivareri:
Note: I have not had a chance to do any extensive testing on this. 

I changed all the places I found that does connection checks, so it should hopefully be consistent across the role.

widhalmt (approved):
Thank you for your work. Even if you missed an occurrence, that's not a problem. Especially since there's more to come. Some connections will be replaced with a module and others should have their own variable (e.g. other roles).

So, approved. Thank you so much.

PR #318, opened 2024-02-12
Add new authors to `AUTHORS`

widhalmt:
Could you please a quick feedback if this is what you wanted? @lcndsmr @ivareri ?


tbauriedel:
We should discuss if you want an automatism for that.

widhalmt:
I have one, more or less - locally. The problem is that the contents of `.mailmap` heavily depend on every contributors personal preferences.

lcndsmr (approved):
Thank you!

PR #323, opened 2024-05-27
replace user and role creation with corresponding modules

christoph2497:
I replaced the old functionality of creating users and roles of logstash with the corresponding modules. Also i added a few new variables to default/main.yml which are used to make the user and role creation more variable. The documenation was also adjusted accordingly. Cause the modules don't support the use of password hashes atm all variables corresponding to password hashes in logstash have been removed.


roles/logstash/tasks/logstash-security.yml
```diff
@@ -378,71 +378,42 @@
...
+          - manage
+          - manage_ilm
+    state: present
+    host: https://localhost:9200
+    auth_user: elastic
+    auth_pass: "{{ logstash_elasticstack_password.stdout }}"
+    verify_certs: false
+    ca_certs: "{{ elasticstack_ca_dir }}/ca.crt"
+
+
```
tbauriedel: double blank line

tbauriedel (changes requested):

roles/logstash/tasks/logstash-security.yml
```diff
@@ -378,71 +378,41 @@
...
+          - logstash_user_indices
+        privileges:
+          - write
+          - create
+          - delete
+          - create_index
+          - manage
+          - manage_ilm
+    state: present
+    host: https://localhost:9200
```
tbauriedel, line 399: should not be hardcoded.
Maybe you can steal something [here](https://github.com/NETWAYS/ansible-collection-elasticstack/blob/main/roles/logstash/tasks/logstash-security.yml#L378)

roles/logstash/tasks/logstash-security.yml
```diff
@@ -378,71 +378,41 @@
...
+  netways.elasticstack.elasticsearch_user:
+    name: "{{ logstash_user }}"
+    fullname: Internal Logstash User
+    password: "{{ logstash_password }}"
+    email: "{{ logstash_email }}"
+    roles:
+      - "{{ logstash_role_name }}"
+    enabled: true
+    state: present
+    host: https://localhost:9200
```
tbauriedel, line 415: should not be hardcoded.
Maybe you can steal something [here](https://github.com/NETWAYS/ansible-collection-elasticstack/blob/main/roles/logstash/tasks/logstash-security.yml#L378)

roles/logstash/tasks/logstash-security.yml
```diff
@@ -378,71 +378,41 @@
...
+          - manage
+          - manage_ilm
+    state: present
+    host: https://localhost:9200
+    auth_user: elastic
+    auth_pass: "{{ logstash_elasticstack_password.stdout }}"
+    verify_certs: false
+    ca_certs: "{{ elasticstack_ca_dir }}/ca.crt"
+
+- name: Create user
```
tbauriedel, line 405: Task name should be more informative

roles/logstash/defaults/main.yml
```diff
@@ -48,8 +48,10 @@ logstash_sniffing: false
 logstash_password_hash: true
 logstash_password_hash_algorithm: bcrypt
 logstash_password_salt_length: 22
+logstash_role_name: logstash_writer
 logstash_user: logstash_writer
 logstash_password: password
+logstash_email: new@user.de
```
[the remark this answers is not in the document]
christoph2497: i will try that

roles/logstash/tasks/logstash-security.yml
```diff
@@ -378,71 +378,41 @@
...
+  netways.elasticstack.elasticsearch_role:
+    name: "{{ logstash_role_name }}"
+    cluster:
+      - manage_index_templates
+      - monitor
+      - manage_ilm
+    indicies:
+      - names:
+          - logstash_user_indices
+        privileges:
```
tbauriedel, line 391: Perhaps we should also consider casting these lines into a variable here.
Same like for "logstash_user_indices"

roles/logstash/tasks/logstash-security.yml
```diff
@@ -378,71 +378,41 @@
...
-    body_format: json
-  register: put_logstash_writer_user_response
-  when: logstash_writer_user_present is not defined
-  delegate_to: "{{ elasticstack_ca }}"
-  run_once: true
-  failed_when: not put_logstash_writer_user_response.json.created
+- name: Create role
+  netways.elasticstack.elasticsearch_role:
+    name: "{{ logstash_role_name }}"
+    cluster:
```
tbauriedel, line 384: Perhaps we should also consider casting these lines into a variable here.
Same like for "logstash_user_indices"

tbauriedel (changes requested):

roles/logstash/tasks/logstash-security.yml
```diff
@@ -378,71 +378,42 @@
...
+- name: Create role
+  netways.elasticstack.elasticsearch_role:
+    name: "{{ logstash_role_name }}"
+    cluster: 
+      - manage_index_templates
+      - monitor
+      - manage_ilm
+    indicies:
+      - names:
+          - logstash_user_indices
```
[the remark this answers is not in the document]
christoph2497: i adopted this from the role jinja template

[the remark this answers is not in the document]
tbauriedel: I don't think it will resolve the argument, it will pass "logstash_user_indices" as a string and distribute it as a permission.
The strings within the variable are then omitted here.

tbauriedel:
Since we want to use these modules and the need the python3 module "elasticsearch", we should consider to install the module as part of the role.

roles/logstash/tasks/logstash-security.yml
christoph2497: fixed with e42504d

roles/logstash/tasks/logstash-security.yml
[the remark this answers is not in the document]
christoph2497: you are right, i have to adjust this that it uses the variable instead

tbauriedel (approved):
LGTM

tbauriedel (approved):

tbauriedel (approved):

tbauriedel (dismissed):

tbauriedel:
Another review is needed because I have added some changes...
Ping @widhalmt

docs/role-logstash.md
```diff
@@ -70,12 +70,13 @@ Aside from `logstash.yml` we can manage Logstashs pipelines.
...
-* *logstash_password_hash_algorithm*: Password hashing algorithms. Value must be same as `xpack.security.authc.password_hashing.algorithm` (default: `bcrypt`)
-* *logstash_password_salt_length*: base64 encoded Salt character lenght. This value must be integer and must be compatible to the selected password hashing algorithms (default: `22`)
-* *logstash_password_hash_salt_seed*: A seed to generate random but idempotent salt on the elasticstack ca host. The salt will be used to create idempotent logstash hashed user password (default: `SeedChangeMe`)
-* *logstash_password*: Password of Elasticsearch user. It must be at least 6 characters long (default: `password`)
-* *logstash_user_indices*: Indices the user has access to (default: `'"ecs-logstash*", "logstash*", "logs*"'`)
+* *logstash_user_email*: email-address that is linked with the logstash_user (Default: `""`)
+* *logstash_user_fullname*: fullname that is linked with the logstash_user (Default: `Internal Logstash User`)
+* *logstash_user_password*: Password of Elasticsearch user. It must be at least 6 characters long (default: `password`)
```
widhalmt, line 75:
```suggestion
* *logstash_user_password*: Password of `logstash_user` in Elasticsearch. It must be at least 6 characters long (default: `password`)
```
I know, this was already wrong before the change. But please use the chance to fix it.

docs/role-logstash.md
```diff
@@ -70,12 +70,13 @@ Aside from `logstash.yml` we can manage Logstashs pipelines.
...
+* *logstash_role_name*: Name of the logstash role that is getting created (Default: `logstash_writer`)
+* *logstash_role_cluster_privileges*: Cluster privileges the role has access to (default: `"manage_index_templates", "monitor", "manage_ilm"`)
+* *logstash_role_indicies_names*: Indices the role has access to (default: `"ecs-logstash*", "logstash*", "logs*"`)
+* *logstash_role_indicies_privileges*: Indices the role has access to (default: `"write", "create", "delete", "create_index", "manage", "manage_ilm"`)
```
widhalmt, line 79:
```suggestion
* *logstash_role_indicies_privileges*: Index permissions the role has on `logstash_role_indicies_names`: (default: `"write", "create", "delete", "create_index", "manage", "manage_ilm"`)
```
I guess, that was just a copy&paste error.

roles/logstash/tasks/logstash-security.yml
```diff
@@ -368,81 +363,32 @@
...
-  failed_when: not put_logstash_writer_user_response.json.created
+- name: Create logstash role {{ logstash_role_name }}
+  netways.elasticstack.elasticsearch_role:
+    name: "{{ logstash_role_name }}"
+    cluster: "{{ logstash_role_cluster_privileges }}"
+    indicies:
+      - names: "{{ logstash_role_indicies_names }}"
+        privileges: "{{ logstash_role_indicies_privileges }}"
+    state: present
+    host: "https://{{ hostvars[elasticstack_ca].ansible_default_ipv4.address }}:{{ elasticstack_elasticsearch_http_port }}"
```
widhalmt, line 374: This will not work on clusters with security disabled. But since we will to have support for them soon, anyway, I'm ok with not creating another version without authentication.

roles/logstash/tasks/logstash-security.yml
```diff
@@ -368,81 +363,32 @@
...
+    auth_user: elastic
+    auth_pass: "{{ logstash_elasticstack_password.stdout }}"
+    verify_certs: false
```
widhalmt, line 377:
```suggestion
    verify_certs: true
```
Please!

roles/logstash/tasks/main.yml
```diff
@@ -11,6 +11,16 @@
   changed_when: false
   when: ansible_os_family == 'Debian'
 
+- name: Install Python Packages
+  ansible.builtin.package:
+    name: ["python3", "python3-pip"]
```
widhalmt, line 16:
```suggestion
    name: 
      - python3
      - python3-pip
```
Matches our other code. And it's easier readable as well as easier maintainable.

Are you sure, this will work on all supported operating systems? I thought, package names are different.

widhalmt (changes requested):
I found a view things that *might* need to be changed.

Please make sure installing the packagaes really works on all operating systems. And maybe move the installation to the `elasticstack` role so we don't have to run the same code every time we need to handle users.

roles/logstash/tasks/main.yml
tbauriedel: Well, all checks have passed. 
If we have checks for all supported systems it should be fine. If not, we will have to build checks for the missing one.

docs/role-logstash.md
tbauriedel: Since the user name isnt hardcoded anymore I would not mention the previous hardcoded username in the doc. I would go with something like "Password of logstash user in Elasticsearch".
Mentioning the `logstash_user` inside of the doc could be misleading if you dont use the default name.

tbauriedel:
Pipelines failing to bug #326

tbauriedel:
(Ansible verbosity 3 needs to be removed before merging)

widhalmt (approved):
Looks good to me. No objections from my side.

PR #326, opened 2024-06-06
Fix minor problems with lint etc.

widhalmt:
With updated versions of tools we use for testing our tests now fail. This PR sums up minor changes to remove lint and other minor flaws.

### DIFF
--- a/.github/workflows/test_full_stack.yml
+++ b/.github/workflows/test_full_stack.yml
@@ -33,7 +33,7 @@ jobs:
       fail-fast: false
       matrix:
         distro:
-          - rockylinux8
+          #- rockylinux8
           - rockylinux9
           - ubuntu2004
           - ubuntu2204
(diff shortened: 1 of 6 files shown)


widhalmt:
Looks like we hit a bug: https://discuss.elastic.co/t/2024-06-06t0032-277-error-o-e-b-elasticsearch-server-domain-co-uk-fatal-exception-while-booting-elasticsearch-java-lang-nullpointerexception-cannot-invoke-org-elasticsearch-nativeaccess-systemd-notify-ready-because-this-systemd-is/360895/6

widhalmt:
Please read the comments above. It seems like there's a bug in ES 8.14.0 that breaks our tests for Rocky Linux.

tbauriedel:
We should create an issue to reenable these checks again, once the bug is fixed.

tbauriedel (approved):

PR #331, opened 2024-07-25
Allow to set a password for Redis connections

### LINKED ISSUE #330: [Feature]: Redis with password
### Describe the feature request

Some environments enforce all services (including Redis) to have passwords.

I don't want to discuss if this is useful or not. But it's a fact that some environment enforce this no matter what. So I will add a way to add a Redis password to every Redis input and output that we're managing.

widhalmt:
I hijacked the logstash_pipelines molecule check to test these new settings.

fixes #330


tbauriedel (approved):
LGTM

PR #336, opened 2024-08-08
Fix Python module installation

### LINKED ISSUE #334: [Bug]: Pip installation of modules fails on some systems

_No response_

### Describe the bug

I used this collection on several installations without problems. But yesterday I ran into a problem I haven't seen before. The collection had a problem to install missing modules via `pip`. Here's the error output:

```
TASK [netways.elasticstack.elasticstack : Install Elasticsearch Python Module] **************************************************************************************************************************************************************
fatal: [elastic01.xxx.at]: FAILED! => {"changed": false, "cmd": ["/usr/bin/python3", "-m", "pip.__main__", "install", "elasticsearch"], "msg": "
:stderr: error: externally-managed-environment

× This environment is externally managed
╰─> To install Python packages system-wide, try apt install
    python3-xyz, where xyz is the package you are trying to\n    install.
    
    If you wish to install a non-Debian-packaged Python package,
    create a virtual environment using python3 -m venv path/to/venv.
    Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
    sure you have python3-full installed.
    
    If you wish to install a non-Debian packaged Python application,
    it may be easiest to use pipx install xyz, which will manage a
    virtual environment for you. Make sure you have pipx installed.
    
    See /usr/share/doc/python3.11/README.venv for more information.

note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
hint: See PEP 668 for the detailed specification.
"}
```
Since this is a customers system it will take a while to get more information if possible at all.

### How to recreate the bug?

I can only assume that there is no `pip` installed on this system. I will try to verify it but for now it's my best guess.

widhalmt:
There are several ways of installing Python modules on a system. We can't make safe assumptions about the system where the collection is used.

Possible problems:
* Modules available via package manager can be too old
* Python could be entirely managed via package manager and pip wouldn't break that
* Pip could be installed in different ways

This change will do the following:

* Install `pip` via package manager if a new variable is set
* Try to install modules via pip
* If that fails because Python is managed via package manager, try installing them via package manager
* One of the above methods should be successful but you still might end up with modules which are too old to work
* If that's the case you can force the installation via pip and break the managed Python installation

fixes #334

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ Every role is documented with all variables, please refer to the documentation f
 
 ## Global variables
 
-* `elasticstack_force_pip`: Will force installation of required Python modules via `pip`. This is useful if your package manager doesn't provide current versions of modules. (Default: `false`)
+* `elasticstack_force_pip`: Will force installation of required Python modules via `pip`. This is useful if your package manager doesn't provide current versions of modules. (Default: `false`) See [PEP668](https://peps.python.org/pep-0668/) for more details.
 * `elasticstack_manage_pip`: Will install `pip` on your system. (Default: `false`)
 
 ## Installation
--- a/README.md
+++ b/README.md
@@ -21,6 +21,11 @@ Every role is documented with all variables, please refer to the documentation f
 * [elasticsearch_role](docs/module-elasticsearch_role.md)
 * [elasticsearch_user](docs/module-elasticsearch_user.md)
 
+## Global variables
+
+* `elasticstack_force_pip`: Will force installation of required Python modules via `pip`. This is useful if your package manager doesn't provide current versions of modules. (Default: `false`)
+* `elasticstack_manage_pip`: Will install `pip` on your system. (Default: `false`)
+
 ## Installation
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -21,6 +21,11 @@ Every role is documented with all variables, please refer to the documentation f
 * [elasticsearch_role](docs/module-elasticsearch_role.md)
 * [elasticsearch_user](docs/module-elasticsearch_user.md)
 
+## Global variables
+
+* `elasticstack_force_pip`: Will force installation of required Python modules via `pip`. This is useful if your package manager doesn't provide current versions of modules. (Default: `false`)
```
martialblog, line 26: Maybe have a reference for future people to PEP 668, https://peps.python.org/pep-0668/ 

Otherwise :+1:

widhalmt: Excellent idea! Just added it (if I didn't introduce another typo, that is)

tbauriedel (approved):

Issue #333, opened 2024-08-02
[Feature]: metricbeat oss > 7


cbeaujoin-stellar:
### Describe the feature request

In the repo role, the following tasks seems to be wrong:
```
- name: Check for versions
  ansible.builtin.fail:
    msg: "No OSS versions later than 7 are available"
  when:
    - elasticstack_release | int > 7
    - elasticstack_variant == "oss"
```

Because there is an availlable version of metricbeats-oss version > 8
https://www.elastic.co/downloads/beats/metricbeat-oss/

So:
```
monitoring_elasticstack_release: 8
monitoring_elasticstack_version: 8.14.3
```
Should be allowed

widhalmt:
Good catch! Thank you. We'll look into it.

PR #338, opened 2024-08-08
add fail message to roles that are not available as oss version

christoph2497:
To prevent the a strange error message that is not easy to understand for the basic user I added a check that fails for specific roles when the release type and version are not available for the specific role.

### DIFF
--- a/.github/workflows/test_role_repos.yml
+++ b/.github/workflows/test_role_repos.yml
@@ -46,10 +46,10 @@ jobs:
           - rockylinux8
           - rockylinux9
           - ubuntu2004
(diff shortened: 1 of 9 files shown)


widhalmt (changes requested):
Please start products with a capital letter in messages. e.g. "Kibana" instead of kibana.

And please refer to the issue you're trying to fix by using "closing keywords". This has to be in the first description of the PR. When the PR is merged, the Issue is automatically closed.
e.g.
fixes #333

widhalmt (approved):
Thanks for your contribution!

PR #344, opened 2024-08-29
Enable use of uniqe tls passwords pr host

ivareri:
Allow use of unique `elasticsearch_tls_key_passphrase` pr host

(not sure if this is a fix or a feature, but I went with fix as that's the behaviour I expected)


widhalmt:
Hm... it seems like the `hostvars` variant of a variable does not fall back to a group var when it's not set. Shouldn't the current code already be able to use per host passwords?

ivareri:
Hum, my understanding was that `hostvars` should include everything as the host would see it, including group_vars. I guess I'll do some testing. 

The problem I'm trying to fix is caused by the certificate generation loop only running when `inventory_hostname == elasticstack_ca`. The current code therefore uses the `elasticsearch_tls_key_passphrase`  for the ca host when generating the certs. It does however use the proper `elasticsearch_tls_key_passphrase` when configuring the nodes. This leaves a broken install on every node but `elasticstack_ca`, as the password set in the configuration does not match the password used when generating the certificate.

widhalmt:
Thanks! Let's see what the tests are saying. :-)

PR #351, opened 2024-11-05
Fix/schedule for kics

widhalmt:
Introduce 2 new incidents that will trigger a KICS (security scan) run:

* Weekly scheduled by cron
* On every `push`. We can rework that later. But it should be the fastest way of getting results into https://github.com/NETWAYS/ansible-collection-elasticstack/security right now. Which will help with assigning tasks.

This will fix the fact that KICS only runs on Pull Requests but does not work as code scanning utility the way GitHub means it.

### DIFF
--- a/.github/workflows/kics.yml
+++ b/.github/workflows/kics.yml
@@ -14,6 +14,8 @@ on:
           - debug
   pull_request:
   merge_group:
+  schedule:
+    - cron: '15 6 * * 4'
 jobs:
   kics:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


widhalmt:
Looking at the tests that ran I see KICS two times. Once for the push and once for the pull request. What do you think how should we get rid of one of the tests? Just use `push`?

.github/workflows/kics.yml
```diff
@@ -13,7 +13,10 @@ on:
           - warning
           - debug
   pull_request:
+  push:
```
tbauriedel, line 16:
```suggestion
```

Using `pull_requests` should be enough. With `push` **and** `pull_request`, the workflow will be triggered twice in PRs.

tbauriedel (changes requested):

tbauriedel (approved):

PR #353, opened 2024-11-06
Add exception to KICS for http in non-security scenario

widhalmt:
We need to use `http` without `s` for certain scenarios. This commit will introduce an exception so KICS will not complain about it.


afeefghannam89:
Do you want to make the same exeption for another lines with http in this PR? Is it possible to define a "global" exception rule like in lint?

afeefghannam89 (approved):

PR #360, opened 2025-01-23
Remove local `become`

widhalmt:
Havin `become` only in tasks where needed might be good coding style but it breaks tests with Rocky Linux and might have other unpredicted side effects. So I'm removing it.

### DIFF
--- a/roles/elasticsearch/tasks/main.yml
+++ b/roles/elasticsearch/tasks/main.yml
@@ -204,7 +204,6 @@
   when: elasticsearch_jna_workaround | bool
 
 - name: Set jvm heap size
-  become: yes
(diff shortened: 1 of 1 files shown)


tbauriedel (dismissed):
I am not familiar with that part of the collection in detail.
But the changes I can see look okay.

afeefghannam89 (approved):

tbauriedel:
@afeefghannam89 
Have you seen the failing actions [here](https://github.com/NETWAYS/ansible-collection-elasticstack/actions/runs/13010654925/job/36287470923)?

tbauriedel:
I have run the actions again.
Seems like it was just a temporary problem that is now working.

PR #361, opened 2025-01-24
fixed permission gap with updating logstash-security.yml

gpeker19:
add permissions and waiting for check. Findings 24 and 26


widhalmt (approved):
Nice, thank you. This will work!

Needs #360 to be merged before checks can run through

PR #373, opened 2025-03-05
Fix task condition

Saeid-Abadi:


widhalmt (approved):
Good catch! Thank you.

Issue #363, opened 2025-01-26
[Question]: Upgrade implementation vs recommendations in docs


ivareri:
### Ask a question

Hi

I've been meaning to provide feedback on the upgrade functionality for quite some time, but life have gotten in the way. Maybe this should have been multiple issues, and I've might have missed some details or points, but it is what is it. 

Latest [elastic docs](https://www.elastic.co/guide/en/elastic-stack/8.17/upgrading-elasticsearch.html) and the implemented process have some differences: 
- Stopping ML nodes is not implemented.
- Docs says `"cluster.routing.allocation.enable": "primaries"` vs implemented `none`
- Docs recommends upgrading tier-by-tier (frozen-cold-warm-hot)

Things I have observed during testing:
- Wait period for a cluster to return to Green status is not always long enough
- Sometimes cluster never returns to Green status as there are no eligible nodes for the replica shards
- If a node fails, the entire play should abort. Currently it just drops the node that failed, and keeps running for the rest of the nodes. 

Questions:
Is the `"cluster.routing.allocation.enable"` based on earlier recommendations, or is there another reason to choose `none` over `primaries`?

My biggest blocker currently is that the cluster remains in a yellow state when there are replicas with no eligible nodes. The Docs says to proceed with the upgrade in these cases. This means we would have to check `init` and `relo` columns in `_cat/health?v=true`. This might either be trivial or far-from-trivial, not sure to be honest. 

Regarding failing entire play vs node, this might be something in my ansible setup, or something in my playbook. I've not had time to give this a hard look yet. 

Adding a task to start/stop ML nodes should be trivial, I might drop a PR for this if/when I find the time.

ivareri:
Ok, so by replacing this test https://github.com/NETWAYS/ansible-collection-elasticstack/blob/87a7dc6ab0149038eb1af145f149e58ad005e145/roles/elasticsearch/tasks/elasticsearch-rolling-upgrade.yml#L102

with

````
 until:
        - "response.json.relocating_shards == 0"
        - "response.json.initializing_shards == 0"
        - "response.json.status == 'green' or
           response.json.status == 'yellow'"
````

it solved the deadlock with no eligble nodes. It should probably verify this status a couple of times before procceding.

@widhalmt: Do you want PRs for this against main or https://github.com/NETWAYS/ansible-collection-elasticstack/pull/349

widhalmt:
Hi,

[...]

PR #378, opened 2025-03-17
Fix/graceful elasticsearch

ivareri:
Fixes https://github.com/NETWAYS/ansible-collection-elasticstack/issues/363


ivareri:
ahh, did not know json_query filters had a dependency on "jmespath". I'll revert that commit to avoid the dependency.

PR #381, opened 2025-04-08
use a unique name for the elasticsearch ca cert on the ansible controller

zerwes:
avoid mixing certs if we have multiple elasticsearch nodes (standalone or clusters)


zerwes:
gosh darn it - wrong commiter

PR #395, opened 2025-09-23
Review and refactoring of the collection

### LINKED ISSUE #316: [Question]: Rename `full_stack` variable - Please let us know your ideas
### Ask a question

After extensive (offline) discussion while reviewing #314 @tbauriedel and I agreed that `elasticstack_full_stack` is confusingly named.

So calling all developers, users and all others: This variable should be `true` when one role inside the collection can rely on things being managed and named by other roles in the same collection. It's `false` when you want to interfere and manage some parts yourself.

It's easier to explain with an example: The `repos` role in the collection will create `yum` configuration to use the repositories of Elastic. There's a hardcoded name in the configuration. And package installation will use these names in connection with `enablerepo` to enable these very repositories. So `elasticstack_full_stack: true`. 
But if your hosts can't reach Elastics repository servers and you configure your own repository mirror for `yum` to use, you can use arbitrary names. If the package installation then tries to enable the repos by name, the task will fail. So `elasticstack_full_stack: false` because the roles can't rely on the repo name being the hardcoded example from `repos` role.

Now it's your turn: How should we name the variable to cause less confusion?

tbauriedel:
PR to add some refactoring development. Will grow over time.  
Part of #396 

A list of changes can be found below.

* Rename `elasticstack_full_stack` to `elasticstack_collection_managed` #316 68535e3
* Sort documentations 0ccd969
* Centralize package installation in role `repos` 56c5751
* Remove workaround for EL >= 9 that is not needed anymore c5b90a3
* Rename `elasticstack_enable_repos` to `repos_enable_repos` and move into the `repos` role itself 2e7d808
* Introduce `repos_deb_remove_legacy_key` to remove key from the deprecated trusted.gpg keyring 4aa385f
* Streamline yaml format and Ansible comments 79c3009
* Update issue templates to use types instead of tags 62e7463
----

fixes #316


afeefghannam89 (dismissed):
I like the idea, thanks. I think will document the following variable in the documentation PR, will you?
elasticstack_hosts_elasticsearch
elasticstack_hosts_logstash
elasticstack_hosts_kibana

PR #418, opened 2026-02-17
Update OS versions for automated tests

widhalmt:
We had some outdated OS in our CI/CD pipelines. This should bring them to a current status.

PR #421, opened 2026-02-21
Workaround for mismatch in Ansible and Python versions

widhalmt:
Add a workaround for a mismatch in Ansible an Python versions. It keeps us from using `get_url`.

Supersedes #418


README.md
```diff
@@ -78,8 +76,15 @@ We know from personal experience, that the collections work in following combina
 
 * CentOS 7 - Elastic Stack 7
```
afeefghannam89: How about deleting every statement about CentOS 7?

Issue #429, opened 2026-03-19
[Bug]: The right Pipelines don't start after pull request creating


afeefghannam89:
**branches** under **pull_request** means the target branch to whom the pull request will be merged and not the actual feature branch name see [here](https://docs.github.com/de/actions/reference/workflows-and-actions/workflow-syntax#onpull_requestpull_request_targetbranchesbranches-ignore)

PR #430, opened 2026-03-25
Refactoring molecule scenarios and GitHub workflows

afeefghannam89:
Here is a list of issues or tasks to be closed before this PR can be merged in main.

- [x] Stop the running of **Test Collection Roles** and all workflows if the PR is draft. #432 
- [x] Replace OS distribution in small Workflows (roles) from Ubuntu 2204 to Debian 13. #433 
- [x] Remove deprecated scenarios and workflows and merge the duplicated ones. #434 
- [x] test_role_beats.yml doesn't use ansible-core 2.19. #441 
- [x] Extend molecule_full_stack in the test_roles_pr.yml with latest Python und Ansible version. #437  
- [x] Check the running condition of EVERY pipeline if they are correct and practice. #436 . #429 
- [x] Check the sense and functionality every molecule scenario. #439 
- [x] Can we get ride of the second full stack run while merging in main? #440 
- [x] Thinking about the names of workflows. #442 
- [x] Stop the outdated Github pipeline run when pushing new changes in the pull request #462 
- [x] Use GitHub cache for the elastic packages #463


.github/workflows/kics.yml
```diff
@@ -13,14 +13,12 @@ on:
...
-  push:
-    branches:
-      - 'main'
   merge_group:
   schedule:
     - cron: '15 6 * * 4'
 jobs:
-  kics:
+  security_scan:
+    if: github.event.pull_request.draft == false
```
widhalmt, line 26:
```suggestion
    if: github.event_name != 'pull_request' || github.event.pull_request.draft == false
```
To be honest, this code is quite new for me. But with an external analysis I found hints that this would crash on non-pull request triggers because the field is actually not set.

.github/workflows/test_roles_pr.yml
```diff
@@ -75,3 +83,58 @@ jobs:
...
+    env:
+      COLLECTION_NAMESPACE: netways
+      COLLECTION_NAME: elasticstack
+
+    strategy:
+      fail-fast: false
+      matrix:
+        distro:
+          - rockylinux9
+          - ubuntu2204
```
widhalmt, line 146: Do you want to keep `Ubuntu` instead of `Debian` to have a more widespread test regime without running all these Distros through a matrix with the same tests?

roles/elasticsearch/tasks/main.yml
```diff
@@ -121,7 +121,7 @@
   ansible.builtin.set_fact:
     elasticsearch_package: >
       {{
-      'elasticsearch' +
+      ('elasticsearch-oss' if elasticstack_variant == 'oss' else 'elasticsearch') +
```
widhalmt, line 124: Do we really need this change? I thought there's no OSS version of Elasticsearch available anymore in the versions we support.

widhalmt (changes requested):
There's only one real change I want to request and this one I didn't find myself. As far as I understand the change has only to be done in this file. The other workflows can not get into a situation where the field is not existent but would be checked for its contents.

.github/workflows/kics.yml
afeefghannam89: This change came with the #443 . When using ==, GitHub Actions converts both sides to numbers:

null → 0
false → 0
GitHub Actions type coercion: null == false → 0 == 0 → true

Also your condition is doing exactly the same as my condition. The only difference is the writing way. I tested my condition with non-pull request like workflow_dispatch, it was working. you can test it yourself if you want. I did push now an explanation for this line for the future.

.github/workflows/test_roles_pr.yml
afeefghannam89: yes

roles/elasticsearch/tasks/main.yml
afeefghannam89: The change was also a part of #443 . Without it, an OSS installation would still try to install the elasticsearch package instead of elasticsearch-oss. This is incorrect for versions < 8, where a separate elasticsearch-oss package exists. This was a Bug in the collection for the 7.x users. You can remove it in your PR #453 . By the way bevor your PR I would make a release. So the changes history will stay understandable for the users.

.github/workflows/kics.yml
widhalmt: Ok. Your test beats all my "I read somewhere that..." Resolving the conversation.

.github/workflows/test_roles_pr.yml
widhalmt: Good idea. I was unsure if I did remember that correctly, but it seems like I did. Well done.

roles/elasticsearch/tasks/main.yml
widhalmt: Ok, fine with me.
And good idea to have it in the 7-removal PR.
Let's discuss the Release topic elsewhere.

widhalmt (dismissed):
From all that I can say, this is ready to merge. What a beast of an PR - great work!

PR #451, opened 2026-03-27
Remove cron schedule for KICS check.

widhalmt:
We don't really need this check scheduled. It should be sufficient checking PRs. And this is what we do.


afeefghannam89 (dismissed):

afeefghannam89:
Solved in #430

PR #453, opened 2026-03-27
Feature/remove es 7

### LINKED ISSUE #289: Elasticsearch 7 can't work without initial_master_nodes
The check in https://github.com/NETWAYS/ansible-collection-elasticstack/blob/main/roles/elasticsearch/templates/elasticsearch.yml.j2#L31 is wrong at least for Elasticsearch 7. Even a single node cluster needs this to be set.

If I remember correctly, we had to remove it on Elasticsearch 8 single node clusters to not get error messages. So either we live with the message or we introduce yet another check and set it always on 7 and only on multinode cluster on 8.

### LINKED ISSUE #446: [Feature]: Remove the support of elasticstack 7
### Describe the feature request

Here is a potential list for the tasks:
- Elasticsearch security is obligatory in release 8, so no need to disable it
- Remove the support for OSS package ONLY in Elasticsearch and Kibana role. OSS version 8 exist for filebeat and logstash.

widhalmt:
Remove version 7 of Elastic Stack from supported versions.

This also means removing OSS versions for Elasticsearch and Kibana.

closes #289 
closes #446


afeefghannam89 (commented):
Some suggestions ;)

.github/workflows/test_full_stack.yml
```diff
@@ -39,7 +39,6 @@ jobs:
         scenario:
           - elasticstack_default
         release:
-          - 7
```
afeefghannam89: Just for information. We already did the removal of the related tasks to elasticstack  7 version in every workflow [here](https://github.com/NETWAYS/ansible-collection-elasticstack/pull/430).

afeefghannam89 (commented):
Now, my comments should work ;)

PR #462, opened 2026-04-22
Cancel outades pipeline runs

afeefghannam89:
When a new push is made to a pull request, a new pipeline run starts. The already running one is no longer needed and only consumes resources.


widhalmt (approved):
Nice! This should save a big portion on ressources in busy development times.

PR #463, opened 2026-04-24
GitHub cache

### LINKED ISSUE #92: Add caching to GitHub actions
Since our testing actions take very long to run, we should check if we can use caching ( https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows ) to speed them up and use less ressources.

It's clearly an optional enhancement we don't *need* for a specific version. But it would have several benefits.

afeefghannam89:
- Add warm_apt_cache.yml. A Dedicated workflow that pre-downloads Elastic apt packages to the GitHub Actions cache (triggered on push to main, nightly, and manually)
- All test workflows restore the apt cache before running molecule. The version-based cache key updates automatically on new Elastic releases
- Add pip caching (actions/setup-python cache: 'pip') to all workflows
- Mount cached packages into molecule Docker containers via read-only volume; copy .deb files into apt archives in prepare.yml
- Fix #92

### DIFF
--- a/.github/workflows/test_linting.yml
+++ b/.github/workflows/test_linting.yml
@@ -30,6 +30,8 @@ jobs:
         uses: actions/setup-python@v6
         with:
           python-version: "3.11"
+          cache: 'pip'
+          cache-dependency-path: requirements-test.txt
 
       - name: Install test dependencies.
         run: |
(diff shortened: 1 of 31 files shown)


test.md
widhalmt, line 1: I guess this file was committed in error?

widhalmt (changes requested):
The only thing I found was the file that seems to be committed by accident.

test.md
afeefghannam89: Good found, yes you are right. I delete it

PR #466, opened 2026-06-17
Fix Debian ansible_distribution_major_version numeric comparison

mnaier:
This change fixes a type comparison issue in the Ansible condition. Previously, ansible_distribution_major_version was compared as a string, which could lead to incorrect lexical comparisons. The update explicitly converts it to an integer using | int, ensuring a proper numeric comparison when checking if the version is less than or equal to 12.


afeefghannam89:
Thank. LGTM.

PR #467, opened 2026-06-24
Fix/345 readme badge link

### LINKED ISSUE #345: [Bug]: Badge in Readme does not link to pipelines

_No response_

### Describe the bug

We have a badge in our README file that shows that status of the main pipeline. It is a link but it only links to itself. We should move that to the results page of the pipeline

### How to recreate the bug?

n/a

Saeid-Abadi:
Closes #345


Saeid-Abadi:
Since the test_full_stack workflow, the only workflow that runs against main is going to be removed, the badge no longer points to anything useful. The remaining workflows only trigger on pull_request and don't produce default-branch runs, so a badge for them would just show "no status" and be misleading rather than helpful.

afeefghannam89 (approved):

PR #468, opened 2026-06-24
feat: declare minimum ansible-core version

### LINKED ISSUE #435: Check the galaxy.yml
**requires_ansible** and **requires_python** are not set. We should look after another possible changes

Saeid-Abadi:
Add meta/runtime.yml with requires_ansible (>=2.18.0) and note the requirement in the README (ansible-core 2.18 needs Python 3.11+ on the control node). 
Closes #435

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,9 @@ collections:
 
 ### Requirements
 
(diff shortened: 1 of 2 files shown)


afeefghannam89 (approved):
Thanks :)

PR #470, opened 2026-06-24
Support the usage of special character in passwords

### LINKED ISSUE #282: elasticsearch_tls_key_passphrase containing $ not working

afeefghannam89:
Fix #282

Use command with argv to ensure that all special characters are handled correctly, including spaces and empty strings, for every password in this collection. For more info [see](https://docs.ansible.com/projects/ansible/latest/collections/ansible/builtin/command_module.html#parameter-argv) 

argv is already a fully constructed argument list, it is passed directly to execve() without any shlex processing. There is no splitting, quote parsing, or expansion. Every element is delivered exactly as provided, byte for byte. That's why using argv is the correct solution for handling arbitrary special characters, whereas the string-based approach (whether command or shell) must always go through some form of tokenization


widhalmt (approved):
Looks good to me. This should get us rid of this "Umlauts" bug. Thanks!

PR #471, opened 2026-06-25
Make JVM heap configurable via logstash_heap

### LINKED ISSUE #229: Logstash: Handle JVM heap size with a variable
At the moment there is no handling for the JVM heap size parameters (Xms/Xmx) for Logstash installations. The jvm.options file is completely untouched and Logstash is using only 1G of memory. 

There should be the same function as for Elasticsearch to use at least half of the available machine memory or configure the Heap size with a variable.

afeefghannam89:
Fix #229


afeefghannam89:
I have tested it.

PR #511, opened 2026-07-27
logstash: make redis host and TLS configurable for simple pipelines

Saeid-Abadi:
The simple input/output (and default redis) templates hardcoded host => localhost with no TLS option, so pipelines could only reach a co-located Redis in plaintext. Add logstash_redis_input_host, logstash_redis_output_host and logstash_redis_tls (defaults keep the previous behaviour: localhost, no TLS).

### DIFF
--- a/roles/logstash/templates/redis-output.conf.j2
+++ b/roles/logstash/templates/redis-output.conf.j2
@@ -10,7 +10,10 @@ filter {
 {% endif %}
 output {
   redis {
-    host => "localhost"
+    host => "{{ logstash_redis_output_host }}"
+{% if logstash_redis_tls | bool %}
+    ssl_enabled => true
+{% endif %}
     data_type => "list"
     key => "input"
 {% if logstash_beats_input_congestion is defined %}    congestion_threshold => {{ logstash_beats_input_congestion }}{% endif %}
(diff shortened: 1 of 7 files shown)


roles/logstash/templates/redis-output.conf.j2
```diff
@@ -10,7 +10,10 @@ filter {
 {% endif %}
 output {
   redis {
-    host => "localhost"
+    host => "{{ logstash_redis_output_host }}"
+{% if logstash_redis_tls | bool %}
+    ssl_enabled => true
```
afeefghannam89, line 15: It would be helpful to supply another SSL options like key und certificate, you can take a look here https://www.elastic.co/docs/reference/logstash/plugins/plugins-outputs-redis#plugins-outputs-redis-ssl_certificate

roles/logstash/templates/simple-output.conf.j2
```diff
@@ -16,7 +16,10 @@ output {
 {% if not loop.first and pipelinename.exclusive | bool %}else {% endif %}{% if (not loop.last and pipelinename.exclusive | bool) or not pipelinename.exclusive %}if {% endif %}{% if output.condition is defined %}{{ output.condition }} {% endif %}{
 {% endif %}
   redis {
-    host => "localhost"
+    host => "{{ logstash_redis_output_host }}"
+{% if logstash_redis_tls | bool %}
+    ssl_enabled => true
```
afeefghannam89, line 21: Same as above.

afeefghannam89 (changes requested):

afeefghannam89 (approved):
Thanks a lot :)

PR #515, opened 2026-07-29
kibana: manage file logging with variables

### LINKED ISSUE #142: Kibana does not log with version 8
Kibana systemd file does not set logging option with a bath, that is why Kibana does not log under **/var/log/kibana**, should we supply logging capability in kibana.yml? What do you think @widhalmt?

Saeid-Abadi:
Kibana previously only logged to the journal and its logging could only be configured through the verbatim kibana_extra_config passthrough.

Add kibana_manage_logging (off by default, so existing behaviour is kept) plus variables for path, file name, level, layout, console output and rotation. The settings are rendered as flat dotted keys, matching both the rest of kibana.yml.j2 and the style of Elastic's shipped config/kibana.yml.

Rotation is handled by Kibana's rolling-file appender, so no logrotate is needed. The role also creates the log directory for the kibana user, which the passthrough could not do.

Closes #142

### DIFF
--- a/roles/kibana/README.md
+++ b/roles/kibana/README.md
@@ -26,6 +26,31 @@ trust and the `kibana_system` credentials — using the Elasticsearch CA.
     - kibana
 ```
 
+By default Kibana only logs to the journal, readable with
+`journalctl -u kibana`. Set `kibana_manage_logging` to write a log file that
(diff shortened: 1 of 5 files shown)


afeefghannam89 (approved):
Thanks for the fix :)

PR #525, opened 2026-08-05
logstash: delete rotated log files at rollover

Saeid-Abadi:
The managed log4j2 configuration rotated files but never deleted them, so /var/log/logstash grew indefinitely (one .gz per day, per appender, since first deployment). The DefaultRolloverStrategy max attribute is not enough here because filePattern combines %d{yyyy-MM-dd} with %i, which restarts the counter on every new day.

Add a Delete action to every RollingFile appender (plain, json, slowlog plain, slowlog json, deprecation and the per-pipeline routing appender), filtered by the appender's own filename glob and IfLastModified. This restores the behaviour Elastic ships for the plain and json appenders and extends it to the appenders where the vendor default is also incomplete.

The retention is controlled by logstash_logging_retention, default 7D, matching Elastic. Any log4j2 IfLastModified duration works, e.g. 24H or P30D.


afeefghannam89 (approved):
Thanks

